PR #66, opened 2019-09-30
BAU: Improve documentation for thumbprint headers.

mrwilson:
The `x5t` and `x5t#256` headers are optional in the RFC but mandatory for the DCS, as we use them to confirm that the certificate used to sign/encrypt is part of a safe-list on the DCS.


bravegrape:
@mrwilson I've solved the merge conflict. There's one thing that stuck out and I want to check with you. The PR now removes this requirement from the message structure:

```
All JWS Headers must include the `x5t` header parameter. This is to allow the DCS to identify the correct certificate to use to check that the message was sent by a specific entity.
```

It's correct to remove this, right? 👀

mrwilson:
@bravegrape - yeah, that's been replaced with explicit lines about `x5t`/`x5t#s256` in both the signing and encryption sub-sections.

PR #73, opened 2019-10-01
Improve wording around forenames parameter

Krenair:


source/api-reference.html.md.erb
```diff
@@ -130,7 +130,7 @@ The number of the passport being checked, which is an integer between 1 and 8999
 String of maximum 30 characters containing the surname as on the passport.
 
 #### forenames
-String of maximum 30 characters, including spaces between forenames. The forenames must be the same as on the passport.
+List of strings of maximum 30 characters, including spaces between forenames. The forenames must be the same as on the passport.
```
Krenair:
```suggestion
List of strings. When concatenated together (including spaces between items) totalling of maximum 30 characters. The forenames must be the same as on the passport.
```

bravegrape: That's remarkably close to what @bethwalton and I came up with: 

List of strings containing the forenames on the passport. The concatenated strings, including spaces between forenames, must be maximum 30 characters.

Krenair: I prefer your one.

Krenair:
```suggestion
List of strings containing the forenames on the passport. The concatenated strings, including spaces between forenames, must be maximum 30 characters.
```

bravegrape: If you commit that version then, I can approve an away we go! 🚀

PR #77, opened 2019-10-07
BAU: Add successful status check example

### LINKED ISSUE #68: Give example for successful DCS status check
The DCS status check call has examples of responses when the DCS is not available for whatever reason.

We should open with the example of the DCS being available.

https://dcs-pilot-docs.cloudapps.digital/api-reference/#response

mrwilson:
Resolves #68


mrwilson:
LGTM!

Issue #76, opened 2019-10-03
Incomplete HTTP response codes docs?


Krenair:
Without reading into the existing DCS code, https://dcs-pilot-docs.cloudapps.digital/status-response-codes/#status-response-codes feels like it might be incomplete? Like there must be a bunch of cases where we can send HTTP 4xx other than 404/429, right? Wrong MIME type, unrecognised x5t header, broken outer signature, broken encryption, broken inner signature, missing or invalid parameter...
`Likewise, an HTTP 200 does not necessarily mean that the DCS is operational.` seems strange, if DCS were unavailable wouldn't we serve HTTP 503?

PR #85, opened 2019-10-22
Review stale pages.

Todai88:
Manual was borking.

### DIFF
--- a/source/status-response-codes.html.md.erb
+++ b/source/status-response-codes.html.md.erb
@@ -14,7 +14,7 @@ The Document Checking Service (DCS) uses standard [HTTP response code][HTTP-Stat
 | 200                | The DCS handled your request successfully. |
 | 404                | The DCS URI is incorrect.              |
 | 429                | You have exceeded your allocated number of API calls.  |
-| 503                | The requested government service is experiencing a planned outage |
+| 503                | The requested government service is experiencing a planned outage. |
 | 5xx                | There is an internal server error. |
 
 Note that the `HTTP 200` response code means the DCS handled your request successfully. It does not confirm whether the passport is valid or not. Likewise, an `HTTP 200` does not necessarily mean that the DCS is operational.
--- a/source/status-response-codes.html.md.erb
+++ b/source/status-response-codes.html.md.erb
@@ -14,7 +14,7 @@ The Document Checking Service (DCS) uses standard [HTTP response code][HTTP-Stat
 | 200                | The DCS handled your request successfully. |
 | 404                | The DCS URI is incorrect.              |
 | 429                | You have exceeded your allocated number of API calls.  |
-| 503                | The requested government service is experiencing a planned outage. |
+| 503                | HM Passport Office is having a planned outage. |
 | 5xx                | There is an internal server error. |
 
 Note that the `HTTP 200` response code means the DCS handled your request successfully. It does not confirm whether the passport is valid or not. Likewise, an `HTTP 200` does not necessarily mean that the DCS is operational.
--- a/source/status-response-codes.html.md.erb
+++ b/source/status-response-codes.html.md.erb
@@ -1,7 +1,7 @@
 ---
 title: Status response codes
 weight: 6
-last_reviewed_on: 2019-10-01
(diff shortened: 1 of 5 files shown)


bravegrape:
So we think the content on those pages is ok to stay as is for another 3 weeks?

Todai88:
> So we think the content on those pages is ok to stay as is for another 3 weeks?

Yes, except for the change regarding the status code, which Daniel has a card for (which has been added in the PR).

source/status-response-codes.html.md.erb
```diff
@@ -14,6 +14,7 @@ The Document Checking Service (DCS) uses standard [HTTP response code][HTTP-Stat
 | 200                | The DCS handled your request successfully. |
 | 404                | The DCS URI is incorrect.              |
 | 429                | You have exceeded your allocated number of API calls.  |
+| 503                | The requested government service is experiencing a planned outage |
```
bjgill:
```suggestion
| 503                | The requested government service is experiencing a planned outage. |
```

source/status-response-codes.html.md.erb
```diff
@@ -14,6 +14,7 @@ The Document Checking Service (DCS) uses standard [HTTP response code][HTTP-Stat
 | 200                | The DCS handled your request successfully. |
 | 404                | The DCS URI is incorrect.              |
 | 429                | You have exceeded your allocated number of API calls.  |
+| 503                | The requested government service is experiencing a planned outage. |
```
bravegrape: In the context of the pilot, it will always be HM Passport Office, right?

If we need to refer to them in generic terms, 'government data source' is more accurate when talking about DCS - as opposed to 'government service'. What do you think?

@bjgill @Todai88 

```suggestion
| 503                | HM Passport Office is having a planned outage. |
```

source/status-response-codes.html.md.erb
```diff
@@ -14,6 +14,7 @@ The Document Checking Service (DCS) uses standard [HTTP response code][HTTP-Stat
 | 200                | The DCS handled your request successfully. |
 | 404                | The DCS URI is incorrect.              |
 | 429                | You have exceeded your allocated number of API calls.  |
+| 503                | HM Passport Office is having a planned outage. |
 | 5xx                | There is an internal server error. |
 
 Note that the `HTTP 200` response code means the DCS handled your request successfully. It does not confirm whether the passport is valid or not. Likewise, an `HTTP 200` does not necessarily mean that the DCS is operational.
```
bravegrape: As




```suggestion
Note that the `HTTP 200` response code means the DCS handled your request successfully. It does not confirm whether the passport is valid or not.
```
As [@Krenair pointed out](https://github.com/alphagov/dcs-pilot-docs/issues/76), if the DCS is not operational for some reason, we'd serve a 503, right?

bjgill: No - DCS does errors in a _fun_ way.

- If DCS has failed catastrophically, we'll return a `500`
- If the outage is planned - `503`
- If DCs is up, but the data source is down (and maybe some other scenarios as well???), we'll return a `200` with a body containing `"error": true` and some error details.

Todai88: I'm re-installing a lot of my software, but from the top of my head (if @dbes-gds can confirm maybe):

* 200 are returned if the request has been handled successfully, regardless of the passport is valid or not
* 5xx is returned in case anything but dmz-proxy is unavailable, or if something unexpected happened
* 503 happens if a government service (hmpo in the context of the pilot), is experiencing a planned outage
* 404 if dmz-proxy doesn't exist, or is down 

> if the DCS is not operational for some reason

In other words, it _depends on the scenario_, or what went wrong. 

A 500 would be served in case something unexpected happened (an app is down, the service is down, something completely un-parsable was sent by the client).

Responding with a 503 is a way for the DCS to allow the client(s) to discern if something planned has gone wrong.

Krenair: Sending HTTP 200 in the event of an upstream API failures does not sound like valid behaviour as no actual check was carried out. HTTP 502 may be correct for this case?

PR #89, opened 2019-11-12
Updating last_reviewed

Todai88:
No need to update page content

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -1,8 +1,8 @@
 ---
 title: Document Checking Service pilot
 weight: 1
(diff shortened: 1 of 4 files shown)


bravegrape:
Just wondering - is it worth having a longer `review_in` period? Is 3 weeks too frequent?

bravegrape:
@Todai88 - I spoke to @mrwilson who suggested extending: `review_in: 6 weeks`. What do you think?

Todai88:
> @Todai88 - I spoke to @mrwilson who suggested extending: `review_in: 6 weeks`. What do you think?

Done

PR #90, opened 2019-11-14
BAU: Explicitly specify that all request fields are mandatory.

mrwilson:


bravegrape:
I'm working on some content improvements related to this - I'll push changes in a little bit

bjgill:
Nice. I wonder if we should include the exact regex we use for validating surnames/forenames. That seems as if it might be useful. A future PR, perhaps...

bravegrape:
> Nice. I wonder if we should include the exact regex we use for validating surnames/forenames. That seems as if it might be useful. A future PR, perhaps...

Interesting 🤔 We should look at this in the user research to see if it comes up. I'll log it as an issue for now so we don't forget 👍

PR #96, opened 2019-11-21
BAU: update manuel

Todai88:
## Why

Manuel was complaining. 🐶 

## What

Just updated the last_reviewed key. There doesn't seem to have been any changes to the relevant code.


mrwilson:
I will merge this as it's a minor change and Jo is out of the office today.

PR #103, opened 2020-01-14
Update gems

bravegrape:
## Why

Keeping our 💎 up to date.

## What

Gem updates resulting from the `bundle update` command.

## How to review

Preview should look alright and checks should pass, as normal 😁


Gemfile.lock
```diff
@@ -68,10 +68,10 @@ GEM
     hamster (3.0.0)
       concurrent-ruby (~> 1.0)
     hashie (3.6.0)
-    html-proofer (3.13.0)
+    html-proofer (3.15.0)
       addressable (~> 2.3)
       mercenary (~> 0.3)
-      nokogiri (~> 1.10)
+      nokogumbo (~> 2.0)
```
bjgill: Is this expected? Otherwise looks fine.

bravegrape: Had a chat with @nickcolley who pointed out that it's `html-proofer` that decided to use `nokogumbo`: https://github.com/gjtorikian/html-proofer/blob/master/html-proofer.gemspec#L23

PR #109, opened 2020-02-05
Add section on rate limits

benjamin-mortimer:
## Why

Pilot participants should be aware there is a rate limit for DCS and what the limit is.

## What

Adds a short section to the API reference page describing the rate limit and the HTTP status codes users receive in the event it is exceeded.


benjamin-mortimer:
Good point @bjgill - I've addressed with some changes. What do you think?

PR #114, opened 2020-02-21
BAU: remove references to 'clientId' from the docs

bjgill:
## Why

We do not use this field for anything - we just record it in our logs. This field is optional. Thus, we can omit it from the pilot docs. This should save pilot participants the effort of providing the field, and us the effort of needing to provide them with the value.

## What

Remove all mentions of the client ID from the docs.


bjgill:
(draft because it's not yet true that the field is optional)

bjgill:
It has now been made optional, so this can be reviewed and then merged

PR #117, opened 2020-02-26
Add a guide to setting up mutual TLS to DCS

benjamin-mortimer:
## Why

Setting up mTLS to DCS was one of the things users struggled with in the lab. We decided to address this with a detailed step by step guide.

## What

Adds a new guide to the docs that walks you through setting up an mTLS connection to DCS, including commands you can use and sample responses.

Also bumps the page weights so it appears in the right place in the navigation. We're going to add some more guides so I've just bumped everything up by 10 points to make room.

## How to review

Please check for factual accuracy.

NB: This guide should contain a link to another guide about raising certificates, which we haven't created yet. I've left it pointing to `/` and created a story in the backlog to update it when the real guide exists.

### DIFF
--- a/source/set-up-mtls-to-dcs.html.md.erb
+++ b/source/set-up-mtls-to-dcs.html.md.erb
@@ -1,0 +1,111 @@
+---
+title: Set up a mutual TLS connection to the DCS
+weight: 3
+last_reviewed_on: 2020-02-26
+review_in: 6 weeks
+---
+
+# Set up a mutual TLS connection to the DCS
+
+You need to use mutual Transport Layer Security (mTLS) to send requests to the DCS. 
+
+mTLS is where the DCS and your service present certificates to each other, so each is sure who the other is. This extra layer of authentication means the DCS can be confident that it only sends information to authorised clients.
+
+Your client will need to present an mTLS certificate to the DCS, which is signed by the GDS Certificate Authority (CA). 
+
+mTLS is also sometimes referred to as ‘client certificate authentication’ or ‘X509 client certificate authentication’.
+
+## Before you start
+
+You’ll need to [generate a private key and request a GDS-signed certificate](/) before you can establish an mTLS connection. It can take up to 2 days for the certificate to be issued.
+
+You’ll also need the DCS CA bundle. Request it from the [DCS pilot helpdesk](/support) if you haven’t received it already.
+
+## 1. Check you have basic connectivity to the DCS
+
+Make sure you have network access to the DCS (for example, you might not have network access if you’re blocked by a firewall).
+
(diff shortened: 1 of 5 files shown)


source/set-up-mtls-to-dcs.html.md.erb
```diff
@@ -0,0 +1,111 @@
...
+
+mTLS is where the DCS and your service present certificates to each other, so each is sure who the other is. This extra layer of authentication means the DCS can be confident that it only sends information to authorised clients.
+
+Your client will need to present an mTLS certificate to the DCS, which is signed by the GDS Certificate Authority (CA). 
+
+mTLS is also sometimes referred to as ‘client certificate authentication’ or ‘X509 client certificate authentication’.
+
+## Before you start
+
+You’ll need to [generate a private key and request a GDS-signed certificate](/) before you can establish an mTLS connection. It can take up to 2 days for the certificate to be issued.
```
bjgill: I assume this is an intentional placeholder, since we have not yet written this page?

source/set-up-mtls-to-dcs.html.md.erb
```diff
@@ -0,0 +1,111 @@
...
+
+You’ll also need the DCS CA bundle. Request it from the [DCS pilot helpdesk](/support) if you haven’t received it already.
```
bjgill: Could we include the CA bundle in the pilot docs? It's just a set of public certificates, so not exactly secret...

source/set-up-mtls-to-dcs.html.md.erb
```diff
@@ -0,0 +1,111 @@
...
+Make an HTTP GET request to: 
+
+```[DCS URL]/checks/passport```
+
+You should get back a response with an HTTP status of 200 and a JSON body as follows: 
+
+```
+{
+“available”: true,
+“scheduleOutages”: []
```
dbes-gds: scheduledOutages

source/set-up-mtls-to-dcs.html.md.erb
bjgill: (never mind - just noticed you've already answered this)

source/set-up-mtls-to-dcs.html.md.erb
benjamin-mortimer: Nice spot - have fixed.

source/set-up-mtls-to-dcs.html.md.erb
benjamin-mortimer: I think we will eventually. However, there are some decisions still pending about it outside the team, so for now I'm leaving as something the user requests.

benjamin-mortimer:
Ta - merging now

PR #120, opened 2020-02-26
DOCATHON: Add generating keys and certs page

chrisclayson:
## Why

Documentation needs more detailed instruction and how to generate keys and obtain required certificates.

## What

Add new page describing end to end process for generating a private
keys, CSRs and submitting to CA.
Update links to this page on other pages.
Describe the changes you're proposing.

### DIFF
--- a/source/generating-keys-and-csrs.html.md.erb
+++ b/source/generating-keys-and-csrs.html.md.erb
@@ -15,16 +15,17 @@ To send requests to DCS you will need to generate separate keys and certificate 
 
 Each environment (integration or production) will require a unique set of keys and certificates for each of these purposes. The certificates must be signed and issued by GDS’ Certificate Authority (CA).
 
-For each purpose, you will need to generate a private key, create a Certificate Signing Request (CSR) and submit it to the GDS CA who will issue you with the certificate. Please note, the issuing process can take up to 2 working days.
+For each purpose, you will need to generate a private key, create a Certificate Signing Request (CSR) and submit the CSR to the GDS CA. The GDS CA will then issue your certificate, please note, the issuing
+process can take up to 2 working days.
 
-We give examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer, however, please ensure you use the naming conventions stated below when raising your CSR.
+We give examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer, however, you must use the naming conventions stated below when raising your CSR.
 
 ## 1. Generate a private key
 
 Generate a private key using OpenSSL:
 
 ```
-$ openssl genrsa -out <private key filename>.pem 4096
+$ openssl genrsa -out <private key filename>.pem 2048
 ```
 
 Check that the key file has been written to your current directory.
@@ -82,9 +83,9 @@ Once you’ve filled in the information, the CSR will be created. Check that the
 
 ## 3. Submit the CSR to the Certificate Authority
 
-Note, there are separate CAs (with different trust chains) for each environment, when submitting requests, please ensure you submit to the correct CA. Certificates issued by the  integration CA will not work in production, and vice versa.
+Note, there are separate CAs (with different trust chains) for each environment. When submitting requests, please ensure you submit to the correct CA. Certificates issued by the integration CA will not work in production, and vice versa.
 
-The [helpdesk](/support) will provide you with the URL to correct CA for your environment.
+The [helpdesk](/support) will provide you with the URL for the correct CA for your environment.
 
 To submit the CSR:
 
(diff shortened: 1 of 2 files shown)


source/generating-keys-and-csrs.html.md.erb
```diff
@@ -0,0 +1,112 @@
...
+
+1. Mutual TLS Client Authentication (TLS)
+1. JSON Signing
+1. JSON Encryption
+
+Each environment (integration or production) will require a unique set of keys and certificates for each of these purposes. The certificates must be signed and issued by GDS’ Certificate Authority (CA).
+
+For each purpose, you will need to generate a private key, create a Certificate Signing Request (CSR) and submit it to the GDS CA who will issue you with the certificate. Please note, the issuing process can take up to 2 working days.
+
+We give examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer, however, please ensure you use the naming conventions stated below when raising your CSR.
```
bjgill: > please ensure

Is this a 'should' or a 'must'? From memory, Andrea recommended against adding words to try to be polite.

source/generating-keys-and-csrs.html.md.erb
```diff
@@ -0,0 +1,112 @@
...
+
+To send requests to DCS you will need to generate separate keys and certificate requests for the following purposes:
+
+1. Mutual TLS Client Authentication (TLS)
+1. JSON Signing
+1. JSON Encryption
+
+Each environment (integration or production) will require a unique set of keys and certificates for each of these purposes. The certificates must be signed and issued by GDS’ Certificate Authority (CA).
+
+For each purpose, you will need to generate a private key, create a Certificate Signing Request (CSR) and submit it to the GDS CA who will issue you with the certificate. Please note, the issuing process can take up to 2 working days.
```
bjgill: Nit: I'm not sure this is quite grammatically correct, and 'it' could be ambiguous.

```suggestion
For each purpose, you will need to generate a private key and then create a Certificate Signing Request (CSR). Submit the CSR to the GDS CA who will issue you with the certificate. Please note, the issuing process can take up to 2 working days.
```

source/generating-keys-and-csrs.html.md.erb
```diff
@@ -0,0 +1,112 @@
...
+
+We give examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer, however, please ensure you use the naming conventions stated below when raising your CSR.
+
+## 1. Generate a private key
+
+Generate a private key using OpenSSL:
+
+```
+$ openssl genrsa -out <private key filename>.pem 4096
```
bjgill: Is this key length right? We seem to be using `2048` for the DCS-owned certificates and existing DCS clients.

source/generating-keys-and-csrs.html.md.erb
```diff
@@ -0,0 +1,112 @@
...
+
+```
+Bobs Widgets TLS Integration 2020-02-24
+```
+
+Once you’ve filled in the information, the CSR will be created. Check that the CSR file has been written to your current directory.
+
+## 3. Submit the CSR to the Certificate Authority
+
+Note, there are separate CAs (with different trust chains) for each environment, when submitting requests, please ensure you submit to the correct CA. Certificates issued by the  integration CA will not work in production, and vice versa.
```
bjgill:
```suggestion
Note, there are separate CAs (with different trust chains) for each environment. When submitting requests, please ensure you submit to the correct CA. Certificates issued by the  integration CA will not work in production, and vice versa.
```

Nit: can we shorten some of the long sentences in this PR?

source/generating-keys-and-csrs.html.md.erb
```diff
@@ -0,0 +1,112 @@
...
+
+The [helpdesk](/support) will provide you with the URL to correct CA for your environment.
```
bjgill:
```suggestion
The [helpdesk](/support) will provide you with the URL for the correct CA for your environment.
```

source/generating-keys-and-csrs.html.md.erb
chrisclayson: This might be a question to the floor ... we might be defaulting to 2048, but should we be encouraging bigger keys? I don't think there is any technical reason not to.

bjgill: Apparently NIST suggest that 2048 is sufficient for now. We only need to be better if we want to be secure after 2030, which is long after the current pilot will have finished. There is a cost in processing time of using larger keys.

https://en.wikipedia.org/wiki/Key_size#Asymmetric_algorithm_key_lengths

PR #121, opened 2020-02-26
mXpQwRlt: merge status code page into API reference

bjgill:
## Why

The meaning of the HTTP status codes that the DCS can return are part of our API. Thus, they belong in the API reference. This responds to a pain point identified by our user research participants, who noted that information on DCS responses was spread out.

Also, user research participants were more interested in the passport validity checking than the status.

## What

Add the status code documentation to the API reference. Also re-order the API reference to put passport validity checking above the status.

## How to review

Review commit by commit - the overall diff will not be very useful due to the reordering of the page.


source/api-reference.html.md.erb
```diff
@@ -1,88 +1,20 @@
...
 weight: 16
-last_reviewed_on: 2020-02-07
+last_reviewed_on: 2020-02-26
 review_in: 3 weeks
 ---
 
 # API reference
 
-The Document Checking Service (DCS) API is based on [REST principles][REST]. It returns data in [JSON format][JSON], and uses standard [HTTP status response codes][status-response-codes].
+The Document Checking Service (DCS) API is based on [REST principles][REST]. It returns data in [JSON format][JSON], and uses standard [HTTP status response codes][HTTP-Status-Code].
```
benjamin-mortimer: Where does `[HTTP-Status-Code]` go?

source/api-reference.html.md.erb
```diff
@@ -140,14 +72,19 @@ The passport expiration date in YYYY-MM-DD format.
...
+| ------------------ | -------------------------------------- |
+| 200                | The DCS handled your request. You must decrypt the response body to discover the outcome of your passport check. |
+| 403                | You have exceeded your total allocated number of API calls. |
+| 404                | The DCS URI is incorrect.              |
+| 429                | You have exceeded your rate limit. |
+| 503                | The DCS is temporarily unavailable because of high traffic. |
 
-* `200` [HTTP response code][status-response-codes] for successful requests
-* `503` [HTTP response code][status-response-codes] if HMPO is having a planned outage
+The DCS returns a `200` HTTP response code with signed and encrypted responses. You must decrypt and verify the body to find out whether the passport was valid, the passport was invalid, or an error occurred.
```
benjamin-mortimer: This line is slightly ambiguous. Would it be correct to say something more like:
"Whenever the DCS returns a 200 response code, the response itself will be signed and encrypted."

source/api-reference.html.md.erb
```diff
@@ -140,14 +72,19 @@ The passport expiration date in YYYY-MM-DD format.
 
 ### Response
 
-A successful request is one that completes without error, regardless of whether the passport is valid or not.
+The table shows the meaning of the HTTP response codes in DCS responses. The DCS may return [other HTTP response codes][HTTP-Status-Code], which have their standard meaning. Non-`200` responses may include a message in the body that provides more explanation of the error.
```
benjamin-mortimer: Could this line be more definite about the other codes? "The DCS may return..." part makes this reference appear to be incomplete. Is it true to say that if DCS returns a code other than the ones listed here, then the code has its standard meaning?

benjamin-mortimer: "Non-`200` responses may include a message in the body that provides more explanation of the error."
This sentence is a bit confusing (is a non-200 always an error?) - does it need to be here?

source/api-reference.html.md.erb
```diff
@@ -203,6 +134,84 @@ Boolean indicating if the passport is valid or not. This field is not included i
 #### errorMessage
 A list of strings which describes the error. This field is present if there was an error processing the request.
 
+## Check the DCS status
+
+You can check the status of the DCS by making a HTTP GET request to the service endpoint.
+
+We recommend you use this endpoint to build appropriate levels of reliability into your service.
```
benjamin-mortimer: Nit: we should avoid saying "we", as it's never entirely clear who "we" refers to.

source/api-reference.html.md.erb
bjgill: https://www.rfc-editor.org/rfc/rfc2616.html#section-10

source/api-reference.html.md.erb
bjgill:
```suggestion
Whenever the DCS returns a `200` HTTP response code, the response body will be signed and encrypted. You must decrypt and verify the body to find out whether the passport was valid, the passport was invalid, or an error occurred.
```

Good idea.

source/api-reference.html.md.erb
bjgill: 👍 I'll shorten the first to make it more definitive. I'll move the second after the table, as that seems to make more sense.

source/api-reference.html.md.erb
bjgill: 👍

bjgill:
Just pushed a commit to handle the comments from @benjamin-mortimer

PR #123, opened 2020-02-27
Add a "How DCS works" overview guide

benjamin-mortimer:
## Why

As someone connecting to DCS
I need to understand how DCS works at a high level
So that I have sufficient context to plan how to integrate my service with it

## What

Combines overview content (including diagrams) from the "Message structure" and "Message flow" guides into a single page.

Deletes the more in-detail content from those guides as they will be superseded by our new step by step guides.

## How to review

This will break some links. I propose coming back to those in a separate PR - shout if that's not cool.


bjgill:
Change looks reasonable. Travis will insist you fix the links, though...

benjamin-mortimer:
Thanks Benjamin, I will appease Travis...

PR #124, opened 2020-02-27
A88pU8RK - Handling DCS reponses

dbes-gds:
## Why

The actual message is wrapped-up in the so-called bento box.  Getting it out is difficult.

## What

Step-by-step guide to handling responses from the DCS.

### DIFF
--- a/source/handling-responses.html.md.erb
+++ b/source/handling-responses.html.md.erb
@@ -7,7 +7,7 @@ review_in: 6 weeks
 
 # Handling responses from the DCS
 
-If you have submitted a technically correct [passport validity check request][check-passport] then the DCS will respond with an HTTP 200 status code (unless some kind of error occurs). As the response body is not in plain text you will need to process it in order to retrieve the human-readable content.  This is done by [unwrapping][cryptography-wrapper] the JSON object in the response: 
+When the DCS responds with a [`200` HTTP response code][check-passport-response], the response body is signed, encrypted, and signed again. To learn the outcome of your passport check, you must process the body to get the plain text response object. This process involves unwrapping the signature and encryption wrapper:
 
 1. Validate the outer signature.
 2. Decrypt the contents.
--- a/source/handling-responses.html.md.erb
+++ b/source/handling-responses.html.md.erb
@@ -1,0 +1,42 @@
+---
+title: Handling responses from the DCS
+weight: 8
+last_reviewed_on: 2020-02-27
+review_in: 6 weeks
+---
+
+# Handling responses from the DCS
+
+If you have submitted a technically correct [passport validity check request][check-passport] then the DCS will respond with an HTTP 200 status code (unless some kind of error occurs). As the response body is not in plain text you will need to process it in order to retrieve the human-readable content.  This is done by [unwrapping][cryptography-wrapper] the JSON object in the response: 
+
+1. Validate the outer signature.
+2. Decrypt the contents.
+3. Validate the inner signature to reveal the message inside. 
+
+You will make use of your library’s JWS signature validation and JWE decryption methods to do this.
+
+## Prerequisites
+
+You should have successfully submitted a request to the DCS which returns HTTP 200.  You will need both the DCS signing certificate we provided and your own encryption private key.
+
+## 1. Retrieve the response body
+
+Using your HTTP library, extract the response body from the DCS http response in a form that can be consumed by your JWS library, such as string. 
+
+## 2. Validate the outer signature
+
+Use your JWS library to validate (remove) the outer signature by passing it the response string along with the DCS signing certificate.  Keep this string with the signature removed for the next step.
+
+## 3. Decrypt the contents
+
+Use your JWE library to decrypt the string from step 2, using your own encryption private key.  Keep this decrypted string for the next step.
+
+## 4. Validate the inner signature
+
+Use your JWS library again to validate (remove) the inner signature by passing it the string from step 3, again using the DCS signing certificate.  Keep this final output for the next step.
+
+## 5. Review and consume the human-readable response
+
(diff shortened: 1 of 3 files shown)


source/handling-responses.html.md.erb
```diff
@@ -0,0 +1,42 @@
+---
+title: Handling responses from the DCS
+weight: 8
+last_reviewed_on: 2020-02-27
+review_in: 6 weeks
+---
+
+# Handling responses from the DCS
+
+If you have submitted a technically correct [passport validity check request][check-passport] then the DCS will respond with an HTTP 200 status code (unless some kind of error occurs). As the response body is not in plain text you will need to process it in order to retrieve the human-readable content.  This is done by [unwrapping][cryptography-wrapper] the JSON object in the response:
```
bjgill: Possibly worth linking to the API reference so they can see what the plain text response body looks like and what response codes can be expected?

Do we also want to link to this page from the API reference?

bjgill: > unwrapping

What is the purpose of this link?

source/handling-responses.html.md.erb
```diff
@@ -0,0 +1,42 @@
...
+
+Using your http library, extract the response body from the DCS http response. The response body needs to be of a data type that can be consumed by your JWS library; likely a string. 
+
+## 2. Validate the outer signature
+
+Use your JWS library to validate (remove) the outer signature by passing it the response string along with the DCS signing certificate.  Keep this string with the signature removed for the next step.
+
+## 3. Decrypt the contents
+
+Use your JWE library to decrypt the string from step 2, making use of your own encryption private key.  Keep this decrypted string for the next step.
```
bjgill:
```suggestion
Use your JWE library to decrypt the string from step 2, using your own encryption private key.  Keep this decrypted string for the next step.
```

Nit: possible simplification of the wording?

source/handling-responses.html.md.erb
```diff
@@ -0,0 +1,42 @@
...
+
+## 4. Validate the inner signature
+
+Use your JWS library again to validate (remove) the inner signature by passing it the string from step 3, again using the DCS signing certificate.  Keep this final output for the next step.
+
+## 5. Review and consume the human-readable response
+
+If all has gone well you will have one of these example [responses][check-passport-response].  Note that because DCS returns `HTTP 200` does not mean that the passport is actually valid.  You will need to look at the contents of the response body itself to find out.  Convert the json document in the response to the format required by your application.
```
bjgill: > If all has gone well you will have one of these example [responses]

Is this true? You will have something that _looks like_ one of the examples.

Also, do we need to repeat information here that is already present in the API reference?

source/partials/_links.erb
```diff
@@ -38,6 +38,8 @@
 [check-passport]: /api-reference/#check-passport-validity
 [check-status]: /api-reference/#check-the-dcs-status
 [connect]: /integrating-with-the-dcs/#integrating-with-the-document-checking-service
+[cryptography]: /message-structure/
```
bjgill: Is the title of this page still correct? It now seems more like a guide for encrypting/signing for DCS.

source/handling-responses.html.md.erb
```diff
@@ -0,0 +1,42 @@
...
+
+You will make use of your library’s JWS signature validation and JWE decryption methods to do this.
+
+## Before you start
+
+You should have successfully submitted a request to the DCS which returns HTTP 200.  You will need both the DCS signing certificate we provided and your own encryption private key.
+
+## 1. Retrieve the response body
+
+Using your http library, extract the response body from the DCS http response. The response body needs to be of a data type that can be consumed by your JWS library; likely a string.
```
bjgill: > http

Should this be capitalised?

bjgill: > The response body needs to be of a data type that can be consumed by your JWS library; likely a string. 

From the gov.uk style guide:

> Do not use semicolons as they are often mis-read. Long sentences using semicolons should be broken up into separate sentences instead.

source/handling-responses.html.md.erb
bjgill: > a technically correct

This is a new term I've not seen elsewhere in our docs. What does it mean?

source/handling-responses.html.md.erb
dbes-gds: updated

source/handling-responses.html.md.erb
dbes-gds: updated

source/handling-responses.html.md.erb
dbes-gds: Can you suggest some changes to this as it's been through a couple of versions?

source/handling-responses.html.md.erb
dbes-gds: Please suggest some changes

source/partials/_links.erb
dbes-gds: I think the title should change and become part of an overview, but best leave that until all the new content is done so we can see what needs updated.

source/handling-responses.html.md.erb
bjgill:
```suggestion
You now have a plain text JSON object for your application to use. [Parse the object][check-passport-response] to discover whether the password was valid, the password was invalid, or an error occurred.
```

This perhaps? It's briefer and conveys the same information?

source/handling-responses.html.md.erb
bjgill:
```suggestion
When the DCS responds with a [`200` HTTP response code][check-passport-response], the response body is signed, encrypted, and signed again. To learn the outcome of your passport check, you must process the body to get the plain text response object. This process involves unwrapping the signature and encryption wrapper:
```

Something like this, perhaps? I've tried to make it clear when and why this processing is required - the link allows users to discover when this processing is not needed.

PR #126, opened 2020-02-28
lK5rICbK: Add prerequisites for integration

Todai88:
## Why

Previously we didn't have a clear definition of what was required before a participant started to create their clients.

## What

This commit adds such a list, with focus on outlining what certificates and keys are required before starting their onboarding process. It also clearly defines which keypairs are provided (by us) and which ones they will have to manually create themselves (or submit CSRs).

I've also added a section to the support section to help define what we expect to see from a pilot participant when they are asking for help.

## How to review

`./preview-with-docker.sh`

### DIFF
--- a/source/integrating-with-the-dcs.html.md.erb
+++ b/source/integrating-with-the-dcs.html.md.erb
@@ -1,7 +1,7 @@
 ---
 title: Integrating with the Document Checking Service
 weight: 2
-last_reviewed_on: 2020-02-24
+last_reviewed_on: 2020-02-28
 review_in: 6 weeks
 ---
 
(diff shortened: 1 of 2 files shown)


source/integrating-with-the-dcs.html.md.erb
```diff
@@ -17,16 +17,65 @@ It's up to you to decide which programming language your client is using. Your c
...
+|  2  | Encryption  | Certificate | DCS supplies | Used to encrypt your request payloads   |
+|  3  | mTLS        | CA Bundle   | DCS supplies | Chain of trust for your mTLS configuration |
+
+
+## How to request your public certificates
+
+The DCS requires your public certificates to be signed by our CA.
+
+You will need access to our Certificate Authority portals in order to submit a Certificate Signing Request. 
+There is one portal for each environment you need to raise CSRs with.
```
mrwilson: Should we add "You will be provided with the portal link ... (etc)" here, like we do with environment links above?

Todai88: Good point, I had that in a previous iteration, but must have gotten rid of that in favour of making the text more concise.

Todai88: @mrwilson Moved things around a bit, should be clearer

source/integrating-with-the-dcs.html.md.erb
```diff
@@ -17,16 +17,63 @@ It's up to you to decide which programming language your client is using. Your c
...
-* receive the DCS's signing and encryption certificates
+### DCS URI
+There is one URI for the production environment, and one URI for the integration environment. These will be provided when your pilot application is approved.
 
-Details about how to raise a certificate signing request with our Certificate Authority will follow after the expression of interest stage of the DCS pilot.
+### Access to DCS Certificate Authority portals
+You will be provided URIs to access to our Certificate Authority portals in order to submit a Certificate Signing Request. 
+There is one portal for each environment you need to raise CSRs with.
+
+Each DCS environment has its own Certificate Authority portal where you can submit your Certificate Signing Request (CSR).
```
mrwilson: Sorry to be annoying, but now that this is no longer the first time we use the term `Certificate Signing Request`, the abbreviation should be on line 28, and we should use the acronym here, I think.

source/integrating-with-the-dcs.html.md.erb
```diff
@@ -17,16 +17,63 @@ It's up to you to decide which programming language your client is using. Your c
...
+
+### Certificates and Keys
+There are two sets of certificates and keys that you need to call the DCS:
+
+As mentioned in previous sections, there are two environments that you will need to integrate with:
+
+* the DCS integration environment
+* the DCS production environment
+
+Each environment will require their own keys and certificates mentioned below.
```
kerrr:
```suggestion
Each environment requires a set of the following keys and certificates.
```

source/integrating-with-the-dcs.html.md.erb
Todai88: 🤦‍♂

source/integrating-with-the-dcs.html.md.erb
```diff
@@ -17,16 +17,63 @@ It's up to you to decide which programming language your client is using. Your c
...
+These certificates will be provided to you when your pilot application has been approved:
+
+| No. |   Purpose   |     Type    |    Origin    |           Usage               |
+|-----|-------------|-------------|--------------|-------------------------------|
+|  1  | Signing     | Certificate | DCS supplies | Used to validate DCS request signatures |
+|  2  | Encryption  | Certificate | DCS supplies | Used to encrypt your request payloads   |
+|  3  | mTLS        | CA Bundle   | DCS supplies | Chain of trust for your mTLS configuration |
+
+
+## How to request your public certificates
```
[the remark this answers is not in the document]
Todai88: I get your point, @kerrr 

However, my reasoning is that I _really_ think it's important to have the "how to" showing up in the sidebar. The way our docs are set up, we only show h1-h2 elements in the sidebar... 

Happy to revisit if you strongly disagree. 😄

PR #127, opened 2020-02-28
YC3GlILa - Add json-schema for passport check request payload JSON to API reference 

dbes-gds:
## Why

The json-schema provided can be used to validate json payloads are structured and formatted correctly.

## What

Add json-schema for passport check request payload JSON to API reference with links to validators.

### DIFF
--- a/source/code/dcs-passport-check-request-payload.json-schema.json
+++ b/source/code/dcs-passport-check-request-payload.json-schema.json
@@ -1,0 +1,65 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://dcs-pilot-docs.cloudapps.digital/code/dcs-check-passport-request-payload.json-schema.json",
+  "title": "Document Checking Service - Passport Check Request Payload JSON Schema",
+  "description": "Use this JSON Schema to validate the request payload json for a passport check before signing and encryption is applied.",
+  "type": "object",
+  "properties": {
+    "correlationId": {
+      "description": "String containing an RFC 4122 compliant Universally Unique Identifier (UUID) which ties together multiple requests in the same session.",
+      "type": "string",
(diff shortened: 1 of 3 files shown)


source/code/dcs-passport-check-request-payload.json-schema.json
```diff
@@ -0,0 +1,66 @@
...
+    "requestId": {
+      "description": "String containing an RFC 4122 compliant UUID which identifies a single request within a session.",
+      "type": "string",
+      "pattern": "^[0-9a-fA-F]{8}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{4}\\-[0-9a-fA-F]{12}$"
+    },
+    "timestamp": {
+      "description": "String in ISO 8601 format which identifies when the request took place.",
+      "type": "string",
+      "format": "date-time",
+      "pattern": "^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])T(2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9]\\.[0-9][0-9][0-9]Z$"
```
bjgill: Is this correct? This is for a date-time in UTC ('Z'). DCS also supports (and some IDPs current send) timestamps with a UTC offset...

dbes-gds: This is the format in EvidenceDTO:

    @NotBlank
    @DateFormat(format = "yyyy-MM-dd'T'HH:mm:ss.SSSZ")
    private final String timestamp;

dbes-gds: though stub-reg appears to accept 2020-02-28T16:03:30.45+01:00

dbes-gds: removed the regex as discussed.

PR #129, opened 2020-02-28
lK5rICbK: Add prerequisites for integration

Todai88:
## Why
Previously we didn't have a clear definition of what was required before a participant started to create their clients.

## What
This commit adds such a list, with focus on outlining what certificates and keys are required before starting their onboarding process. It also clearly defines which keypairs are provided (by us) and which ones they will have to manually create themselves (or submit CSRs).

I've also added a section to the support section to help define what we expect to see from a pilot participant when they are asking for help.

## How to review
`./preview-with-docker.sh`

### DIFF
--- a/source/integrating-with-the-dcs.html.md.erb
+++ b/source/integrating-with-the-dcs.html.md.erb
@@ -1,8 +1,7 @@
 ---
 title: Integrating with the Document Checking Service
 weight: 3
-last_reviewed_on: 2020-02-24
-review_in: 6 weeks
+last_reviewed_on: 2020-02-28
 ---
 
 # Integrating with the Document Checking Service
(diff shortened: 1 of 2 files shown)


mrwilson:
Hey Jo - content is great, looks like there are still some conflicts, but would you mind cleaning up the history a lil' bit before merge with some squashing?

(https://medium.com/nona-web/cleaning-up-commit-history-with-git-rebase-61dc41d2829c was what we tried to do on the other project earlier today)

Todai88:
> Hey Jo - content is great, looks like there are still some conflicts, but would you mind cleaning up the history a lil' bit before merge with some squashing?
> 
> (https://medium.com/nona-web/cleaning-up-commit-history-with-git-rebase-61dc41d2829c was what we tried to do on the other project earlier today)

Yep yep, gimme a mo

PR #130, opened 2020-02-28
Remove "Integrating..." section and put all overview content on the homepage

benjamin-mortimer:
## Why

The "Integrating with DCS" guide provided some high-level information about the process you'd follow to connect up to DCS for the pilot. Most of that is covered in more detail with the new guides we've written since the lab day.

## What

+ deletes the "Integrating with DCS" guide 
+ retains some of the content in simplified form and moves it to the homepage
+ remove some content from the homepage and simplify what's left

## How to review

I don't think I've deleted something crucial, but might be worth a quick check.

I think I'll need help with some merge conflicts too...

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -1,22 +1,41 @@
 ---
 title: Document Checking Service pilot
 weight: 1
-last_reviewed_on: 2020-02-24
+last_reviewed_on: 2020-02-28
 review_in: 6 weeks
 ---
 
 # Document Checking Service pilot
 
-This technical documentation is for organisations taking part in the [Document Checking Service (DCS) pilot][govuk-link]. The pilot will allow organisations to digitally check the validity of British passports.
+This technical documentation is for organisations taking part in the Document Checking Service (DCS) pilot. 
 
-Use this documentation to find out:
+You can use the DCS to check digitally whether a given British passport is valid or not.
 
-* [how the DCS works][message-flow]
-* what is involved in [integrating with the DCS API][connect]
-* how to [structure requests for the DCS API][message-structure]
-* [what you can do with the DCS API][api-reference]
+GDS will onboard selected pilot participants during a ‘connecting window’ starting in early 2020. 
 
-We will onboard selected pilot participants during a 'connecting window' starting from the beginning of 2020.
-We will be refining the onboarding process as we go, and participants should therefore be aware that the process may change between iterations.
+## Integrating with the DCS API
+
+To integrate your service with the DCS API you will need to:
+
+1. Develop a client to integrate with the DCS API.
+1. Test your client in a dedicated DCS test environment.
+1. Run your client in the production environment as part of the DCS pilot.
+
+It's up to you to decide which programming language your client uses.
+
+This documentation explains [how the DCS works] and provides step by step guidance on how to:
+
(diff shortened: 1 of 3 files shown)


source/index.html.md.erb
```diff
@@ -1,22 +1,41 @@
...
 review_in: 6 weeks
 ---
 
 # Document Checking Service pilot
 
-This technical documentation is for organisations taking part in the [Document Checking Service (DCS) pilot][govuk-link]. The pilot will allow organisations to digitally check the validity of British passports.
+This technical documentation is for organisations taking part in the Document Checking Service (DCS) pilot. 
 
-Use this documentation to find out:
+You can use the DCS to check digitally whether a given British passport is valid or not.
```
bjgill: I think it's a UK passport, rather than a British passport.

```suggestion
You can use the DCS to check digitally whether a given UK passport is valid or not.
```

source/index.html.md.erb
```diff
@@ -1,22 +1,41 @@
...
 
-* [how the DCS works][message-flow]
-* what is involved in [integrating with the DCS API][connect]
-* how to [structure requests for the DCS API][message-structure]
-* [what you can do with the DCS API][api-reference]
+GDS will onboard selected pilot participants during a ‘connecting window’ starting in early 2020.
```
bjgill: > in early 2020

It's now the end of February. Is it still really 'early 2020'? Maybe 'spring 2020', since that's similarly vague and possibly a bit more accurate?

source/index.html.md.erb
```diff
@@ -1,22 +1,41 @@
...
+1. Run your client in the production environment as part of the DCS pilot.
+
+It's up to you to decide which programming language your client uses.
+
+This documentation explains [how the DCS works] and provides step by step guidance on how to:
+
++ generate keys and request certificates for your DCS connection
++ set up a mutual TLS connection to DCS
++ sign and encrypt a DCS payload
++ handle responses from the DCS
```
bjgill: Would it help users to turn these into links?

source/index.html.md.erb
```diff
@@ -1,22 +1,41 @@
...
+
+To integrate your service with the DCS API you will need to:
+
+1. Develop a client to integrate with the DCS API.
+1. Test your client in a dedicated DCS test environment.
+1. Run your client in the production environment as part of the DCS pilot.
+
+It's up to you to decide which programming language your client uses.
+
+This documentation explains [how the DCS works] and provides step by step guidance on how to:
```
bjgill: Just checked, and this link seems to be a bit broken...

source/index.html.md.erb
benjamin-mortimer: Fair point - have gone with 'spring' now

source/index.html.md.erb
benjamin-mortimer: Agreed - I've made them all links except 'sign and encrypt', as the guide for that has not been published yet

I've noted we need to revisit this in https://trello.com/c/nCDU5YO3/237-link-to-sign-and-encrypt-guide-from-homepage-once-the-guide-is-published

source/index.html.md.erb
benjamin-mortimer: Fixed

PR #138, opened 2020-03-09
Be more up front and specific about certificate naming conventions

benjamin-mortimer:
## Why

Certificate names are used both by the participants and by the DCS team to manage the participant's connection in various ways. Having clear, predictable certificate names makes this much easier. The best way to establish this among multiple parties is with a clearly explained naming convention.

## What

Copies out the certificate naming section from this guide and puts it before the rest of the steps. Left it in its original place too, as don't think the duplication is harmful and may be a useful reminder.


source/generating-keys-and-csrs.html.md.erb
```diff
@@ -15,10 +15,28 @@ To send requests to DCS you will need to generate separate keys and certificate
...
-process can take up to 2 working days.
+For each purpose, you will need to generate a private key, create a Certificate Signing Request (CSR) and submit the CSR to the GDS CA. The GDS CA will then issue your certificate.
+
+It takes up to 2 working days to issue certificates.
+
+The following examples show how to follow this process using OpenSSL. You can use other methods if you prefer.
+
+## Naming your certificates
+
+Follow the DCS certificates naming convention:
```
bjgill:
```suggestion
Follow the DCS certificate naming convention:
```

source/generating-keys-and-csrs.html.md.erb
```diff
@@ -15,10 +15,28 @@ To send requests to DCS you will need to generate separate keys and certificate
...
+
+```
+[Organisation name] [certificate purpose, i.e.  ‘JSON Encryption’, ‘JSON Signing’ or ‘TLS’] [environment, i.e. ‘Integration’ or ‘Production’] [date CSR created]
+```
+
+For example:
+
+    ```
```
bjgill:
```suggestion
```
```

File: source/generating-keys-and-csrs.html.md.erb [hunk 3]
```diff
@@ -15,10 +15,27 @@ To send requests to DCS you will need to generate separate keys and certificate
...
+Follow the DCS certificate naming convention:
+
+```
+[Organisation name] [certificate purpose, i.e.  ‘JSON Encryption’, ‘JSON Signing’ or ‘TLS’] [environment, i.e. ‘Integration’ or ‘Production’] [date CSR created]
+```
+
+For example:
+
+Bobs Widgets TLS Integration 2020-02-24
+```
```
bjgill: Oh dear - it looks as if we've discovered a problem with the suggestions. I think you'll need to add the three back-ticks back in (with the correct lack of indentation).

bjgill: I think you'll need to do it manually - I don't think I can craft a suggestion that will work here...

PR #139, opened 2020-03-16
Generate keys request certificates

PippaClarkGDS:
## Why

Creating a new page so users can understand how to generate keys and request certificates

## What

New page content (needs identified in the UR lab days). Page content has been reviewed by DCS SME and 2 technical writers. If content is changed, would need tech writers to review again. 

## How to review

Content should not need in depth reviewing as it has been reviewed by DCS SME and 2 technical writers. If content is changed, would need tech writers to review again.


source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+weight: 4.3
+last_reviewed_on: 2020-03-09
+review_in: 6 weeks
+---
+
+# Generate keys and request certificates
+
+To send requests to the Document Checking Service (DCS) you’ll need to generate separate keys and certificate requests for:
+
+* mutual TLS Client Authentication (TLS)
```
benjamin-mortimer: The abbreviation should be "mTLS" - plain old "TLS" is a slightly different thing.

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+* JSON signing
+* JSON encryption
+
+For integration, you’ll need to generate your keys and certificates using the following steps.
```
benjamin-mortimer: I suggest losing "for integration" here. 

It's ambiguous (to me) whether "for integration" means "in order to integrate" or "for the integration environment specifically".

If it's the former, I think it's unnecessary as we've already said you need to do this to be able "to send requests".

If the latter, I think it's still unnecessary as the guide makes no further reference to environment-specific things (eg for integration, for production, etc.)

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+1. Generate a private key.
+1. Create a Certificate Signing Request (CSR).
+1. Submit the CSR to the GDS Certificate Authority (CA).
+1. The GDS CA will issue your certificate, which can take up to 2 working days.
+
+The GDS CA must sign and issue the certificates.
+
+This documentation gives examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer.
+## Generate a private key
```
benjamin-mortimer: I think you'll need a line break before this h2 to make the markdown render correctly. Some markdown parsers can handle a missing line break but I'm not sure if the one we use can (there's a few other instances of this in the rest of the doc, I'll flag them).

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+Generate a private key using OpenSSL by entering this code into terminal:
+
+```
+$ openssl genrsa -out <private key filename>.pem 2048
+```
+
+Check you write the key file to your current directory.
+## Create a Certificate Signing Request (CSR)
```
benjamin-mortimer: Line break for h2

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+```
+Bob’s Widgets TLS Integration 2020-02-02
+```
+
+Once you’ve filled in the information, you will have created the CSR.
+
+Check you write the CSR file to your current directory.
+
+## Submit the CSR to the Certificate Authority
```
benjamin-mortimer: I suggest adding "(CA)" here as we use this abbreviation on line 95 and technically haven't explained it.

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+To submit the CSR:
+
+1. Go to the GDS CA - the DCS team will have emailed you a link and you can [contact the helpdesk](/support) if you cannot find the link.
+1. Select __Enrol__.
+1. Upload your CSR file.
+1. Fill in the details on the enrolment form - some details will be pre-populated from data in the CSR, and you should not change these fields.
+1. When filling in the details, make sure you choose the correct ‘Certificate profile’ - choose either __JSON signing__, __JSON encryption__ or __TLS__.
+
+Use the challenge password you chose when you created the CSR.
+## Receive and save your certificates
```
benjamin-mortimer: Line break for h2

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -0,0 +1,120 @@
+---
+title: Message structure
```
bjgill: Is this intentional?

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -0,0 +1,120 @@
...
+1. Submit the CSR to the GDS Certificate Authority (CA).
+1. The GDS CA will issue your certificate, which can take up to 2 working days.
+
+The GDS CA must sign and issue the certificates.
+
+This documentation gives examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer.
+
+## Generate a private key
+
+Generate a private key using OpenSSL by entering this code into terminal:
```
bjgill: Possibly worth being explicit that it needs to be an **RSA** private key? I don't think we support other types of private key.

PippaClarkGDS: I can change this - what does RSA stand for?

source/generate-keys-and-request-certificates.html.md.erb
PippaClarkGDS: oop will change to 'Generate keys and request certificates'

source/generate-keys-and-request-certificates.html.md.erb
PippaClarkGDS: or will a user know what an 'RSA key' is?

benjamin-mortimer: I think so. I also assume that doesn't require any further instructions seeing as the `genrsa` in the code snippet looks like it'll generate an RSA key?

kerrr: RSA stands for Rivest Shamir Adleman, the inventors. But I don't think anyone needs to know that, we can treat RSA as a term of art.

bjgill: Agreed - this was more about being explicit in case a client generates the key by some other means.

PippaClarkGDS: super - thanks 👍

PR #142, opened 2020-03-23
Jose thumbprints

PippaClarkGDS:
## Why

Uploading new content and removing old content to help the docs be more effective and clear in taking the user through the steps needed 

## What

Uploading new content about JOSE thumbprints and signing and encrypting a DCS payload and removing old content which would conflict with new content.

## How to review

These pages have been reviewed by DCS SMEs and 2 tech writers. Any significant changes will need to be 2i'd by a tech writer again. 

Think I fixed all links which would be broken by pages being deleted and links between the 2 new pages but please check these.


source/Set-up-your-JOSE-certificate-thumbprints.html.md.erb
```diff
@@ -0,0 +1,86 @@
...
+xHllK2Kqd5dTQQavfdPYdGIm8w8j6ow/LUPvoCLwwDu+wtgb5Q==
+-----END CERTIFICATE-----
+```
+
+The certificate may have a large block of text describing its contents before the certificate block. You can ignore this.
+
+In the PEM format, the text between `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----` is the Base64 encoding of the binary DER format.
+
+You must then decode the Base64 string to get the binary DER.
+## Hash the binary representation of the certificate
```
benjamin-mortimer: Line break

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+The JWS headers contain information about how you have signed the payload, along with certificate thumbprints which allow the DCS to look up which certificate you’re using.
+
+For the DCS, you need to set the following headers:
+
+1. `x5t` contains the SHA1 thumbprint of your certificate.
+1. `x5t#S256` containing the SHA256 thumbprint of your certificate.
+1. `alg` must be set to `RS256`.
+
+You must set the `x5t` and `x5t#S256` headers. To set them, you’ll need to have [generated thumbprints for your signing certificate](/generate-keys-and-request-certificates/#generate-keys-and-request-certificates).
```
benjamin-mortimer: This should link to the thumbprints guide, rather than the certificates guide.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+}
+```
+## Sign your JSON payload to get a JWS object
+
+You need to sign your JSON payload using a JWS library. To configure your signing of the JWS object in your chosen library, you should use:
+
+* the JWS headers you created
+* the DCS’s client signing key
+
+## Encrypt your JWS object to get a JWE object
```
benjamin-mortimer: In the Signing section, we've separated out the creation of headers as a separate step - I think it would be helpful for users for us to follow the same structure here.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+```
+### Using encryption algorithms
+
+The encrypted component of the JavaScript Object Signing and Encryption (JOSE) message must use:
+
+* `RSA-OAEP` as the algorithm
+* `A128CBC-HS256` as the encryption method
+
+You can [learn more about these in the JSON Web Algorithms specification](https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40).
+## Sign your JWE object to complete the process
```
benjamin-mortimer: Line break

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+You have already signed your JSON payload for the first time before you encrypted the payload.
+
+After encrypting your JWS object, you need to use JWS to sign your payload for a second time.
+
+In your library, you should use the:
+
+* JWS headers you created in the previous step
+* client signing certificate you created in the prerequisites (or in the separate certificate page)
```
benjamin-mortimer: I think this is a typo.

I believe you use the DCS signing certificate for this, as in step one.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+* `RSA-OAEP` as the algorithm
+* `A128CBC-HS256` as the encryption method
+
+You can [learn more about these in the JSON Web Algorithms specification](https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40).
+## Sign your JWE object to complete the process
+
+You have already signed your JSON payload for the first time before you encrypted the payload.
+
+After encrypting your JWS object, you need to use JWS to sign your payload for a second time.
```
benjamin-mortimer: Suggest changing:
After encrypting your JWS object

To:
After creating your JWE object

Because
It's confusing to mention JWS twice in the same sentence to refer to different things

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+
+The encrypted component of the JavaScript Object Signing and Encryption (JOSE) message must use:
+
+* `RSA-OAEP` as the algorithm
+* `A128CBC-HS256` as the encryption method
+
+You can [learn more about these in the JSON Web Algorithms specification](https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40).
+## Sign your JWE object to complete the process
+
+You have already signed your JSON payload for the first time before you encrypted the payload.
```
benjamin-mortimer: Suggest changing:
You have already...

To:
You already...

Because:
The tense is a bit weird in this sentence.

source/support.html.md.erb
```diff
@@ -1,23 +1,18 @@
...
+last_reviewed_on: 2020-03-25
 review_in: 5 weeks
 ---
 
 # Support
 
-## Contact us
+Email <dcs-pilot-support@digital.cabinet-office.gov.uk> if you:
 
-### Security
```
benjamin-mortimer: I'd quite like to keep this as its own header unless there's a reason not to.

It's something users potentially will be scanning for - and if they are scanning for it we definitely want them to find it.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
PippaClarkGDS: sorted - thanks

source/sign-and-encrypt-a-DCS-payload.html.md.erb
PippaClarkGDS: sorted - thanks

source/Set-up-your-JOSE-certificate-thumbprints.html.md.erb
PippaClarkGDS: sorted - thanks

source/sign-and-encrypt-a-DCS-payload.html.md.erb
PippaClarkGDS: ok I can change - what would the new steps look like?

source/sign-and-encrypt-a-DCS-payload.html.md.erb
PippaClarkGDS: nice - cheers Ben

source/sign-and-encrypt-a-DCS-payload.html.md.erb
PippaClarkGDS: changed - thanks

source/support.html.md.erb
PippaClarkGDS: gotcha - a fair point. I've made some amends and included a separate header for 'Security' now

source/sign-and-encrypt-a-DCS-payload.html.md.erb
PippaClarkGDS: no problem - following the same structure is always good. What needs to be changed to follow this?

benjamin-mortimer: On second thoughts, I think it's okay as it is.

PippaClarkGDS: great stuff

source/sign-and-encrypt-a-DCS-payload.html.md.erb
benjamin-mortimer: I think it should just be: "Use the same JWS tools and headers you created when you first signed your JSON payload" - you're just doing the same thing as before.

Worth a developer checking that though.

benjamin-mortimer:
I think it might be the uppercase 'S' in the "Set up thumbprints" file name that's breaking the build.

It's Middleman trying to check the links it builds:

```*  internally linking to /set-up-your-JOSE-certificate-thumbprints/#set-up-your-jose-certificate-thumbprints, which does not exist (line 259)```

It's creating that path lower case, but the file itself has it upper case.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
bjgill: Ben: agreed - that is accurate

PippaClarkGDS: thanks both - will make the change and push

bjgill:
I'll try running this locally to see if I can find the right link to use

PippaClarkGDS:
Created 2 new docs pages about signing and encrypting a DCS payload and JOSE certificate thumbprints and removed 2 pages ('Message flow' and 'Message structure') which would have had duplicate content with the new pages.

PR #143, opened 2020-04-01
Tell users they need to use a JW* library

benjamin-mortimer:
We'd referenced "your library" a few times without explicitly saying you need to use one. This should avoid some confusion.

This PR adds a short section to the 'sign and encrypt' guide to tell users they need to pick a library. We should come back another time with some recommendations for common languages.

It also adds a note to the 'thumbprints' guide to say you don't need to generate them if your library will do it for you, which might save some users some time.

### DIFF
--- a/source/Set-up-your-JOSE-certificate-thumbprints.html.md.erb
+++ b/source/Set-up-your-JOSE-certificate-thumbprints.html.md.erb
@@ -11,9 +11,9 @@ You’ll need to include a certificate ‘thumbprint’ (also sometimes known as
 
 The thumbprint is a hash of the public certificate used to sign or encrypt the JWS or JWE object. DCS uses the thumbprint to look up which certificate to use to validate your signature or decrypt your message.
 
-The [library you use to build JWS and JWE objects](/sign-and-encrypt-a-DCS-payload/#choose-a-library-to-use-for-jws-and-jwe) may be able to create thumbprints for you. If you need to create the thumbprints yourself, follow this guide.
+The [library you use to build JWS and JWE objects](/sign-and-encrypt-a-DCS-payload/#choose-a-library-to-use-for-jws-and-jwe) may be able to create thumbprints for you.
 
-To manually set up your certificate thumbprints, you’ll need to:
+If you need to create the thumbprints manually yourself, you’ll need to:
 
 1. Check your certificate is in binary DER format.
 1. Hash the binary representation of the certificate.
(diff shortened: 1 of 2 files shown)


source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -19,6 +20,10 @@ To build the JSON Web Signature (JWS) and JSON Web Encryption (JWE) wrapper for
 
 The diagram shows the steps you must take to build the signing and encryption wrapper for your JSON object. It shows you start with a plain JSON object. You then sign this object to get a JWS object. You then encrypt this JWS object to get a JWE object. Finally, the diagram shows you sign your JWE object. This will give you a final payload which has been signed, encrypted, and signed again.
 
+## Choose a library to use for JWS and JWE
+
+Use a well-maintained library for your language to help you build JWS and JWE objects.
```
PippaClarkGDS, line 25: wondering if we can give users more direction about where to find these 'well-maintained' libraries?

source/Set-up-your-JOSE-certificate-thumbprints.html.md.erb
```diff
@@ -11,7 +11,9 @@ You’ll need to include a certificate ‘thumbprint’ (also sometimes known as
 
 The thumbprint is a hash of the public certificate used to sign or encrypt the JWS or JWE object. DCS uses the thumbprint to look up which certificate to use to validate your signature or decrypt your message.
 
-To set up your certificate thumbprints, you’ll need to:
+The [library you use to build JWS and JWE objects](/sign-and-encrypt-a-DCS-payload/#choose-a-library-to-use-for-jws-and-jwe) may be able to create thumbprints for you. If you need to create the thumbprints yourself, follow this guide.
```
PippaClarkGDS, line 14: this works v well. I think we can combine the last 2 sentences: 

"If you need to manually create the thumbprints yourself, you'll need to:" (and then go to the numbered list)

source/sign-and-encrypt-a-DCS-payload.html.md.erb
benjamin-mortimer: I'd like to in a later pass at this - we should give a list of recommended libraries.

But we don't have that list at the moment.

PippaClarkGDS: great stuff - sounds good

PR #148, opened 2020-04-15
added 'Access the data in a DCS response' content from gdoc

PippaClarkGDS:
## Why

User needs to know how to access the data in a DCS response

## What

New content on how to access the data in a DCS response

## How to review

2 tech writers have already reviewed - if there's any substantial changes, they will need to review again

### DIFF
--- a/source/access-the-data-in-a-DCS-response.html.md.erb
+++ b/source/access-the-data-in-a-DCS-response.html.md.erb
@@ -1,0 +1,81 @@
+---
+title: Access the data in a DCS response
+weight: 4.7
+last_reviewed_on: 2020-04-20
+review_in: 6 weeks
+---
+
+# Access the data in a DCS response
+
+When you make a successful request to the Document Checking Service (DCS), the DCS will respond with:
+
+* a `200` HTTP response code
+* a [JOSE object](https://dcs-pilot-docs.cloudapps.digital/Set-up-your-JOSE-certificate-thumbprints/#set-up-your-jose-certificate-thumbprints) containing the data you requested
+
+Follow these instructions to access the data in the DCS response. You’ll need to:
+
+1. Retrieve the response body.
+1. Validate the outer signature.
+1. Decrypt the JSON Web Encryption (JWE) object inside.
+1. Validate the inner signature to reveal the JSON message inside.
+
+You can read about the data contained in JSON messages in the [API reference section](https://dcs-pilot-docs.cloudapps.digital/api-reference/#api-reference).
+
+## Prerequisites
+
+You should have successfully submitted a request to the DCS which returns a `200` HTTP response code.
+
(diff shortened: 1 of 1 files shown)


source/access-the-data-in-a-DCS-response.html.md.erb
```diff
@@ -0,0 +1,81 @@
...
+
+You’ll get a JSON object containing the passport information you requested from DCS.
+
+## Review and use the human-readable response
+
+You now have a plain text JSON object for your application to use. Read the object to discover whether:
+
+* the passport was valid
+* the passport was invalid
+an error occurred in a DCS data provider
```
bjgill, line 79: Is this supposed to be part of the list?

source/access-the-data-in-a-DCS-response.html.md.erb
```diff
@@ -0,0 +1,81 @@
...
+last_reviewed_on: 2020-04-20
+review_in: 6 weeks
+---
+
+# Access the data in a DCS response
+
+When you make a successful request to the Document Checking Service (DCS), the DCS will respond with:
+
+* a `200` HTTP response code
+* a [JOSE object](https://dcs-pilot-docs.cloudapps.digital/Set-up-your-JOSE-certificate-thumbprints/#set-up-your-jose-certificate-thumbprints) containing the data you requested
```
bjgill, line 13:
```suggestion
* a [JOSE object](/Set-up-your-JOSE-certificate-thumbprints/#set-up-your-jose-certificate-thumbprints) containing the data you requested
```

For internal links, you may wish to use relative links (as per my suggestion) instead of absolute links (as in your original). That means that the link will always point to the right page, even when the pilot docs are being hosted at another URL (e.g. locally on localhost).

See other pages (e.g. 'Set up your JOSE certificate thumbprints') for examples of how this is done.

(the other links as well)

source/access-the-data-in-a-DCS-response.html.md.erb
```diff
@@ -0,0 +1,81 @@
...
+
+Follow these instructions to access the data in the DCS response. You’ll need to:
+
+1. Retrieve the response body.
+1. Validate the outer signature.
+1. Decrypt the JSON Web Encryption (JWE) object inside.
+1. Validate the inner signature to reveal the JSON message inside.
+
+You can read about the data contained in JSON messages in the [API reference section](https://dcs-pilot-docs.cloudapps.digital/api-reference/#api-reference).
```
bjgill, line 22: Possibly worth linking directly to the section on that page on responses? (`/api-reference/#check-passport-validity-response`)

source/access-the-data-in-a-DCS-response.html.md.erb
PippaClarkGDS: nice spot - have amended

source/access-the-data-in-a-DCS-response.html.md.erb
PippaClarkGDS: ah gotcha - thanks for the info, sounds like a more sensible approach. Will do this in future. Have amended

source/access-the-data-in-a-DCS-response.html.md.erb
PippaClarkGDS: will do

PR #152, opened 2020-04-21
G7RK6f1q: document how quotas work

bjgill:
## Why
Pilot clients have quotas that limit the total number of checks they can make. We need to document how this works and how clients can find out information on their usage.

Note that this only applies to pilot clients - our existing clients do not have quotas.
## What

Explain what quotas are and how they appear for pilot clients. Document the headers clients can use to monitor their usage.

### DIFF
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -220,7 +220,7 @@ In both cases, the [`Retry-After`][Retry-After] HTTP header specifies how many s
 
 Each organisation connected to the DCS for the pilot will have a quota of DCS checks. This is a fixed limit on the number of production DCS checks that the organisation can make during the pilot.
 
-The DCS returns a 403 (Forbidden) HTTP status code if you have used up your quota. You will be able to make no further DCS checks after this point.
+The DCS returns a 403 (Forbidden) HTTP status code if you have used up your quota. You won't be able to make any more DCS checks after this point.
 
 You can see your quota usage in HTTP headers in passport validity responses from the DCS.
 
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -216,4 +216,17 @@ DCS returns a:
 
 In both cases, the [`Retry-After`][Retry-After] HTTP header specifies how many seconds you should wait before retrying.
 
+## Quotas
+
+Each organisation connected to the DCS for the pilot will have a quota of DCS checks. This is a fixed limit on the number of production DCS checks that the organisation can make during the pilot.
+
+The DCS returns a 403 (Forbidden) HTTP status code if you have used up your quota. You will be able to make no further DCS checks after this point.
+
+You can see your quota usage in HTTP headers in passport validity responses from the DCS.
+
+* `DCS-Quota-Limit` is the total number of checks you will be able to make as part of the DCS pilot
+* `DCS-Quota-Remaining` is the number of checks you have remaining
+
+These headers will be present in responses with 200 and 403 HTTP status codes. The DCS may also add these headers to other responses.
+
 <%= partial "partials/links" %>
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -203,7 +203,7 @@ Boolean indicating if the passport is valid or not. This field is not included i
 #### errorMessage
 A list of strings which describes the error. This field is present if there was an error processing the request.
 
-## Rate limits
+## Understand rate limits in the DCS
 
 DCS handles up to 50 requests in total every 10 seconds.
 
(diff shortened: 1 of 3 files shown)


source/api-reference.html.md.erb
```diff
@@ -216,4 +216,17 @@ DCS returns a:
 
 In both cases, the [`Retry-After`][Retry-After] HTTP header specifies how many seconds you should wait before retrying.
 
+## Quotas
+
+Each organisation connected to the DCS for the pilot will have a quota of DCS checks. This is a fixed limit on the number of production DCS checks that the organisation can make during the pilot.
+
+The DCS returns a 403 (Forbidden) HTTP status code if you have used up your quota. You will be able to make no further DCS checks after this point.
```
benjamin-mortimer, line 223: Suggest small style change:

```suggestion
The DCS returns a 403 (Forbidden) HTTP status code if you have used up your quota. You won't be able to make any more DCS checks after this point.
```

source/api-reference.html.md.erb
```diff
@@ -216,4 +216,17 @@ DCS returns a:
 
 In both cases, the [`Retry-After`][Retry-After] HTTP header specifies how many seconds you should wait before retrying.
 
+## Quotas
```
PippaClarkGDS, line 219: could this be made more action-based ... ie what's the action a user will get from this heading? Maybe 'Understanding quotas in the DCS'?

source/api-reference.html.md.erb
```diff
@@ -216,4 +216,17 @@ DCS returns a:
...
+
+Each organisation connected to the DCS for the pilot will have a quota of DCS checks. This is a fixed limit on the number of production DCS checks that the organisation can make during the pilot.
```
PippaClarkGDS, line 221: keeping how we refer to the DCS consistent as we say 'the DCS' in docs. Would suggest: 

"Each organisation connected to the DCS for the pilot will have a quota of checks for the DCS. This quota is a fixed limit on how many checks the organisation can make in DCS production during the pilot."

source/api-reference.html.md.erb
PippaClarkGDS: in the style guide, we don't use 'n't' ... use 'will not' instead. 

Also can we just cut 'any more' as they won't be able to make any DCS checks after this point? It'll make the sentence cleaner

source/api-reference.html.md.erb
```diff
@@ -216,4 +216,17 @@ DCS returns a:
...
+
+The DCS returns a 403 (Forbidden) HTTP status code if you have used up your quota. You will be able to make no further DCS checks after this point.
+
+You can see your quota usage in HTTP headers in passport validity responses from the DCS.
```
PippaClarkGDS, line 225: could we make it clearer as it's not clear how the user gets this? ... something like "You can see your quota usage in  the HTTP headers when you receive passport validity responses from the DCS."?

source/api-reference.html.md.erb
```diff
@@ -216,4 +216,17 @@ DCS returns a:
...
+
+* `DCS-Quota-Limit` is the total number of checks you will be able to make as part of the DCS pilot
```
PippaClarkGDS, line 227: small thing but use 'you'll' instead of 'you will'

source/api-reference.html.md.erb
```diff
@@ -216,4 +216,17 @@ DCS returns a:
...
+* `DCS-Quota-Remaining` is the number of checks you have remaining
+
+These headers will be present in responses with 200 and 403 HTTP status codes. The DCS may also add these headers to other responses.
```
PippaClarkGDS, line 230: what does 'these headers will be present' mean? What are we trying to tell the user here?

source/status-response-codes.html.md.erb
```diff
@@ -12,6 +12,7 @@ The Document Checking Service (DCS) uses standard [HTTP response code][HTTP-Stat
 | HTTP response code | Description                            |
 | ------------------ | -------------------------------------- |
 | 200                | The DCS handled your request successfully. |
+| 403                | You have used up [your quota of DCS checks][quotas] and are no longer permitted to make DCS checks.
```
PippaClarkGDS, line 15: maybe instead of 'no longer permitted to make DCS checks' ... 'and cannot make any more DCS checks'

source/api-reference.html.md.erb
bjgill: I think we're trying to help clients consume the headers by making it clear in which types of responses they can expect to see the headers.

But maybe we could delete this paragraph. I don't think it would surprise clients that the headers were not present in some error responses.

PippaClarkGDS: happy to delete 👍

source/api-reference.html.md.erb
```diff
@@ -216,4 +216,15 @@ DCS returns a:
 
 In both cases, the [`Retry-After`][Retry-After] HTTP header specifies how many seconds you should wait before retrying.
 
+## Understand quotas in the DCS
+
+Each organisation connected to the DCS for the pilot will have a quota of checks for the DCS. This quota is a fixed limit on how many checks the organisation can make in DCS production during the pilot.
+
+The DCS returns a 403 (Forbidden) HTTP status code if you have used up your quota. You won't be able to make any more DCS checks after this point.
```
PippaClarkGDS, line 223: change 'won't' to 'will not'. 

And can we remove the 'any more' DCS checks to make the sentence cleaner/shorter or do we need to keep it in?

bjgill:
```suggestion
The DCS returns a 403 (Forbidden) HTTP status code if you have used up your quota. You will not be able to make DCS checks after this point.
```

Like that.

PR #153, opened 2020-04-24
kvMZlCZz: document JOSE libraries we've tested

bjgill:
## Why

From our user research, we've seen that users appreciate knowing which JOSE libraries are known to work with the DCS.

Be careful with the working to make it clear that we've tested that they work. Do not recommend using any particular library as we have not done enough in-depth investigation of them to check that they are suitable (e.g. secure, well-maintained, licensing).

## What

Document the libraries with which I am certain that we have made successful DCS checks.

I've tested the clients written in Python and Java, so know that they work against the DCS. For the other languages, I've consulted with the people who wrote clients using those languages - see the individual commits for details.

Note that we had a number of user research participants attempt to use Ruby to connect to the DCS. However, they all ran out of time before making a successful check. So we cannot confidently say that any particular Ruby libraries do/don't work with the DCS.

### DIFF
--- a/source/sign-and-encrypt-a-DCS-payload.html.md.erb
+++ b/source/sign-and-encrypt-a-DCS-payload.html.md.erb
@@ -29,6 +29,7 @@ The DCS team have tested that you can make successful DCS checks using these lib
 * C#: [jose-jwt](https://www.nuget.org/packages/jose-jwt/)
 * Go: [jose2go](https://github.com/dvsekhvalnov/jose2go)
 * Java: [Nimbus JOSE + JWT](https://connect2id.com/products/nimbus-jose-jwt)
+* node.js: [jose](https://www.npmjs.com/package/jose)
 * Python: [JWCrypto](https://jwcrypto.readthedocs.io/en/latest/)
 
 ## Build a plain JSON payload
--- a/source/sign-and-encrypt-a-DCS-payload.html.md.erb
+++ b/source/sign-and-encrypt-a-DCS-payload.html.md.erb
@@ -24,6 +24,11 @@ The diagram shows the steps you must take to build the signing and encryption wr
 
 Use a well-maintained library for your language to help you build JWS and JWE objects.
 
+The DCS team have tested that you can make successful DCS checks using these libraries:
+
+* Java: [Nimbus JOSE + JWT](https://connect2id.com/products/nimbus-jose-jwt)
+* Python: [JWCrypto](https://jwcrypto.readthedocs.io/en/latest/)
+
 ## Build a plain JSON payload
 
 Create a JSON object containing the details of the passport you want to check, for example:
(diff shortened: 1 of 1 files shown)


bjgill:
@PippaClarkGDS - I'd appreciate your view as to whether the wording in this PR accurately conveys the intended meaning?

PippaClarkGDS:
> @PippaClarkGDS - I'd appreciate your view as to whether the wording in this PR accurately conveys the intended meaning?

Looks good Benjamin. If there was a longer list of should haves or must nots, I'd recommend a bullet list to get all the information over without losing anything. But this looks good as there's only a couple of points.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -24,6 +24,11 @@ The diagram shows the steps you must take to build the signing and encryption wr
 
 Use a well-maintained library for your language to help you build JWS and JWE objects.
 
+The DCS team have tested that you can make successful DCS checks using these libraries:
+
+* Java: [Nimbus JOSE + JWT](https://connect2id.com/products/nimbus-jose-jwt)
+* Python: [JWCrypto](https://jwcrypto.readthedocs.io/en/latest/)
+
```
chrisclayson, line 31: For .NET both Lewis and I used [`jose-jwt` ](https://www.nuget.org/packages/jose-jwt/)

bjgill:
Thanks - I've added commits adding in the other libraries we've tested. I've also updated the PR description to reflect what's been done.

I think this can now be approved and merged. Ideally, I'd like reviews from @chrisclayson, @Todai88, and @tomrosier to check that I've not made a mistake with the commits that reference them.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -24,6 +24,14 @@ The diagram shows the steps you must take to build the signing and encryption wr
 
 Use a well-maintained library for your language to help you build JWS and JWE objects.
 
+The DCS team have tested that you can make successful DCS checks using these libraries:
+
+* C#: [jose-jwt](https://www.nuget.org/packages/jose-jwt/)
```
chrisclayson, line 29: I would say .NET rather than C#, as the library can be used with any .NET language, not necessarily C#.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -24,6 +24,14 @@ The diagram shows the steps you must take to build the signing and encryption wr
...
+* Go: [jose2go](https://github.com/dvsekhvalnov/jose2go)
```
Todai88, line 30: 👍

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -24,6 +24,14 @@ The diagram shows the steps you must take to build the signing and encryption wr
 
 Use a well-maintained library for your language to help you build JWS and JWE objects.
 
+The DCS team have tested that you can make successful DCS checks using these libraries:
+
+* Go: [jose2go](https://github.com/dvsekhvalnov/jose2go)
+* Java: [Nimbus JOSE + JWT](https://connect2id.com/products/nimbus-jose-jwt)
+* .NET: [jose-jwt](https://www.nuget.org/packages/jose-jwt/)
```
chrisclayson, line 31: 👍

PR #156, opened 2020-04-30
Change connection start date to acknowledge delays

benjamin-mortimer:
Removed the 'connecting window' concept as we don't really use this - we're just going to connect people.

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -9,7 +9,8 @@ review_in: 6 weeks
 
 This technical documentation is for organisations taking part in the [Document Checking Service (DCS) pilot][govuk-link]. The pilot will allow organisations to digitally check the validity of British passports.
 
-We will onboard selected pilot participants during a 'connecting window' starting from the beginning of 2020.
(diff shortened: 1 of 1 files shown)


source/index.html.md.erb
```diff
@@ -9,7 +9,8 @@ review_in: 6 weeks
 
 This technical documentation is for organisations taking part in the [Document Checking Service (DCS) pilot][govuk-link]. The pilot will allow organisations to digitally check the validity of British passports.
 
-We will onboard selected pilot participants during a 'connecting window' starting from the beginning of 2020.
+We will onboard selected pilot participants in early 2020.
```
bjgill, line 12: This is technically true, but I'm not sure May really counts as 'early 2020'. Might "mid-2020" is similar be more accurate?

Also, the current wording might suggest that we will finish onboarding in early 2020. Maybe something like "We will start onboarding ..."

benjamin-mortimer: I agree about "start onboarding", but I think May is just about 'early' - if you consider the year as two halves, for example.

PR #158, opened 2020-04-30
BR9gNJHs: Add information about generating request-specific data in the payload

mrwilson:
## Why

Make it clear that correlationId, requestId, and timestamp need to be generated by clients within their application.

## What

* Add very minor information - almost all programming languages support generating uuids and timestamps in the correct format.
* Fix minor error in _link.erb where the ISO is mis-numbered

## How to review

`./preview-with-docker.sh` and navigate to `http://localhost:4567/sign-and-encrypt-a-DCS-payload/#build-a-plain-json-payload`

### DIFF
--- a/source/sign-and-encrypt-a-DCS-payload.html.md.erb
+++ b/source/sign-and-encrypt-a-DCS-payload.html.md.erb
@@ -59,7 +59,7 @@ You must generate the following fields in your application:
 
 Most programming languages support generating RFC 4122-compliant universally unique identifiers (UUIDs), and ISO 8601-compliant timestamps.
 
-We recommend using [version 4 UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) as they are based on random input.
+We recommend using [version 4 UUIDs](https://tools.ietf.org/html/rfc4122#section-4.4) as they are based on random input.
 
 ## Create JWS headers
 
--- a/source/sign-and-encrypt-a-DCS-payload.html.md.erb
+++ b/source/sign-and-encrypt-a-DCS-payload.html.md.erb
@@ -38,8 +38,8 @@ Create a JSON object containing the details of the passport you want to check, f
 
 ```json
 {
-  "correlationId": "550e8400-e29b-41d4-a716-446655440000",
-  "requestId": "550e8400-e29b-41d4-a716-446655440003",
+  "correlationId": "21206c34-4d97-47c4-8aee-6fb0d80b311b",
+  "requestId": "e1651f51-5db8-4e8c-9cf7-68fd4063404e",
   "timestamp": "1997-07-16T19:20:30.45+01:00",
   "passportNumber": "123456789",
   "surname": "Smith",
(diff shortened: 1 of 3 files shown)


source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -51,6 +51,13 @@ Create a JSON object containing the details of the passport you want to check, f
   "expiryDate": "2020-01-01"
 }
 ```
+
+You must generate `correlationId`, `requestId`, and `timestamp` in your own application.
+
+Most programming languages support generating [RFC 4122] compliant UUIDs and [ISO 8601] compliant timestamps.
+
+We recommend using [version 4 UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) as they are based on random input.
```
bjgill, line 59: Grammar nit: should UUID be plural here?

Otherwise looks sensible.

mrwilson: D'oh! You're absolutely correct.

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -51,6 +51,13 @@ Create a JSON object containing the details of the passport you want to check, f
   "expiryDate": "2020-01-01"
 }
 ```
+
+You must generate `correlationId`, `requestId`, and `timestamp` in your own application.
```
PippaClarkGDS, line 55: might work better as 'You must generate these (codes?) in your own application: 

- `correlationId`
- `requestId`
- `timestamp`

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -51,6 +51,13 @@ Create a JSON object containing the details of the passport you want to check, f
...
+
+Most programming languages support generating [RFC 4122] compliant UUIDs and [ISO 8601] compliant timestamps.
```
PippaClarkGDS, line 57:
```suggestion
Most programming languages support generating RFC 4122-compliant universally unique identifiers (UUIDs), and ISO 8601-compliant timestamps.
```

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -51,6 +51,13 @@ Create a JSON object containing the details of the passport you want to check, f
...
+
+We recommend using [version 4 UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) as they are based on random input.
```
PippaClarkGDS, line 59:
```suggestion
We recommend [using version 4 UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) as they are based on random input.
```

PippaClarkGDS: is there a more reliable/trusted link we can use instead of Wikipedia?

mrwilson: We could use the RFC (https://tools.ietf.org/html/rfc4122#section-4.4) but that's more how to implement generation of UUIDs, rather than what they are.

Perhaps this doesn't require a link at all?

PippaClarkGDS: if there's a better link that's more reliable than Wikipedia, I'd recommend including that. Otherwise, we can stick with Wiki as a source as it would be useful for users to be able to know more about v4 UUIDs

PR #161, opened 2020-05-06
uploaded the gdoc content for sending a test request

PippaClarkGDS:
## Why

currently a gap in the docs. This doc plugs the gap by outlining how users can test their client works by sending test requests. 

## What

Page which explains how a user can send different test requests to the DCS. 

## How to review

Already been reviewed by 2 tech writers. If there's substantial changes, will need to get them to review again.

### DIFF
--- a/source/send-a-test-request-to-the-DCS.html.md.erb
+++ b/source/send-a-test-request-to-the-DCS.html.md.erb
@@ -13,6 +13,9 @@ You’ll now have a client which can:
 
 You must check whether your client can make successful passport validity requests using the test passport data on this page.
 
+Use the DCS integration endpoint for your test checks. [Contact the DCS team](/support) if you need the endpoint URL.
+
+
 ## Generate a `valid true` response
 
 Enter these sample passport details in your client to test a `valid true` response.
--- a/source/send-a-test-request-to-the-DCS.html.md.erb
+++ b/source/send-a-test-request-to-the-DCS.html.md.erb
@@ -1,0 +1,117 @@
+---
+title: Send a test request to the DCS
+weight: 4.65
+last_reviewed_on: 2020-04-20
+review_in: 6 weeks
+---
+
+# Send a test request to the DCS
+You’ll now have a client which can:
+
+* [sign and encrypt a Document Checking Service (DCS) payload](/sign-and-encrypt-a-DCS-payload/#sign-and-encrypt-a-dcs-payload)
+* [make requests secured by mutual Transport Layer Security (mTLS)](/create-a-network-connection/#create-a-network-connection-to-the-dcs) to the DCS
+
+You must check whether your client can make successful passport validity requests using the test passport data on this page.
+
+## Generate a `valid true` response
+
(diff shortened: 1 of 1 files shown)


source/send-a-test-request-to-the-DCS.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+You’ll now have a client which can:
+
+* [sign and encrypt a Document Checking Service (DCS) payload](/sign-and-encrypt-a-DCS-payload/#sign-and-encrypt-a-dcs-payload)
+* [make requests secured by mutual Transport Layer Security (mTLS)](/create-a-network-connection/#create-a-network-connection-to-the-dcs) to the DCS
+
+You must check whether your client can make successful passport validity requests using the test passport data on this page.
+
+## Generate a `valid true` response
+
+Enter these sample passport details in your client to test a `valid true` response.
```
Todai88, line 18: To not risk confusing clients, would it be worth specifying what environment (integration, production) they will target with the request? 

Meaning that the requests outlined in this document would only give the outlined responses when issued against the integration environment? Assuming that's true.

PippaClarkGDS: sounds good to me - nothing wrong with not leaving any assumptions. Where would be best to mention this? Perhaps near the beginning?

PippaClarkGDS: cheers Jo @Todai88 - make sense to me. Where would be best to include this info? At the beginning perhaps? Something like "You must check whether your client can make successful passport validity requests to the integration environment using the test passport data on this page."

source/send-a-test-request-to-the-DCS.html.md.erb
```diff
@@ -0,0 +1,117 @@
...
+---
+
+# Send a test request to the DCS
+You’ll now have a client which can:
+
+* [sign and encrypt a Document Checking Service (DCS) payload](/sign-and-encrypt-a-DCS-payload/#sign-and-encrypt-a-dcs-payload)
+* [make requests secured by mutual Transport Layer Security (mTLS)](/create-a-network-connection/#create-a-network-connection-to-the-dcs) to the DCS
+
+You must check whether your client can make successful passport validity requests using the test passport data on this page.
+
```
benjamin-mortimer, line 15:
```suggestion
Use the DCS integration endpoint for your test checks. [Contact the DCS team](/support) if you need to find out the endpoint URL.
```

source/send-a-test-request-to-the-DCS.html.md.erb
benjamin-mortimer: @PippaClarkGDS I've added a suggested line for where it should go

Todai88: Hi, yeah. What ben said :)

PippaClarkGDS: ta! Have put that in now. Could you both re-review?

PR #167, opened 2020-05-12
9fT1MR50: add integration url

tomrosier:
## Why

Supply the integration url in the documentation.

## What

Adding the integration URL.

### DIFF
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -22,15 +22,13 @@ You can check the status of the DCS by making a HTTP GET request to the service 
 
 We recommend you use this endpoint to build appropriate levels of reliability into your service.
 
-### DCS Endpoints
+### The DCS integration endpoint
 
-Integration endpoint: <https://dcs-integration.ida.digital.cabinet-office.gov.uk/>
+The DCS integration endpoint is <https://dcs-integration.ida.digital.cabinet-office.gov.uk/>.
 
 ### Request
 
-`GET: https://<DCS-URI>/checks/passport`
-
-The `<DCS-URI>` will be supplied at a later date.
+`GET: /checks/passport`
 
 The HTTP header must contain `Content-Type: application/jose`.
 
@@ -94,9 +92,7 @@ You can use the DCS API to check if a passport is valid.
 
 ### Request
 
-`POST: https://<DCS-URI>/checks/passport`
-
-The `<DCS-URI>` will be supplied at a later date.
+`POST: /checks/passport`
 
 The HTTP header must contain `Content-Type: application/jose`.
 
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -22,6 +22,10 @@ You can check the status of the DCS by making a HTTP GET request to the service 
 
 We recommend you use this endpoint to build appropriate levels of reliability into your service.
 
+### DCS Endpoints
+
+Integration endpoint: <https://dcs-integration.ida.digital.cabinet-office.gov.uk/>
+
 ### Request
 
 `GET: https://<DCS-URI>/checks/passport`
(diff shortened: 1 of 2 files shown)


mrwilson:
Hey Tom, thanks for this.

Can we also update the API reference page (https://dcs-pilot-docs.cloudapps.digital/api-reference/#api-reference) with the URL too? There are some placeholders there.

source/send-a-test-request-to-the-DCS.html.md.erb
```diff
@@ -13,7 +13,7 @@ You’ll now have a client which can:
 
 You must check whether your client can make successful passport validity requests using the test passport data on this page.
 
-Use the DCS integration endpoint for your test checks. [Contact the DCS team](/support) if you need the endpoint URL.
+Use the DCS integration endpoint for your test checks. `Integration endpoint: https://dcs-integration.ida.digital.cabinet-office.gov.uk/`
```
bjgill, line 16: Is it intentional that this should not be a rendered as a link?
image

bjgill: Also, possibly worth including the `/checks/passport` at the end? Might prevent a client or two from trying the wrong URL...

mrwilson: We could take inspiration from Notify's docs, I think - https://docs.notifications.service.gov.uk/rest-api.html#base-url

mrwilson:
Sorry for being a bit fiddly, but can we wrap the url to the integration environment in a code block like the Notify docs? (https://docs.notifications.service.gov.uk/rest-api.html#base-url)

I also suggest we tidy up some of the other references in the API docs from:
>`GET: https://<DCS-URI>/checks/passport`
>
> The `<DCS-URI>` will be supplied at a later date.

to

> `GET /checks/passport`

As we provide the URL at the top (the same applies for the POST request)

source/api-reference.html.md.erb
```diff
@@ -22,6 +22,10 @@ You can check the status of the DCS by making a HTTP GET request to the service
 
 We recommend you use this endpoint to build appropriate levels of reliability into your service.
 
+### DCS Endpoints
```
PippaClarkGDS, line 25:
```suggestion
### The DCS integration endpoint
```

source/api-reference.html.md.erb
```diff
@@ -22,6 +22,10 @@ You can check the status of the DCS by making a HTTP GET request to the service
...
+
+Integration endpoint: <https://dcs-integration.ida.digital.cabinet-office.gov.uk/>
```
PippaClarkGDS, line 27:
```suggestion
The DCS integration endpoint is <https://dcs-integration.ida.digital.cabinet-office.gov.uk/>.
```

source/send-a-test-request-to-the-DCS.html.md.erb
```diff
@@ -13,7 +13,7 @@ You’ll now have a client which can:
 
 You must check whether your client can make successful passport validity requests using the test passport data on this page.
 
-Use the DCS integration endpoint for your test checks. [Contact the DCS team](/support) if you need the endpoint URL.
+Use the DCS integration endpoint for your test checks. Integration endpoint: <https://dcs-integration.ida.digital.cabinet-office.gov.uk/checks/passport>
```
PippaClarkGDS, line 16: can we embed the hyperlink in the 'Use the DCS integration endpoint' bit of text?

mrwilson: For normal links to other pieces of web content, this would be appropriate, but navigating to the DCS url in a browser doesn't really make sense as a user journey (their request would immediately fail). I think we're good with this way?

PR #168, opened 2020-05-12
BAU: update guidance on algorithms

bjgill:
## Why

Be more permissive where possible to make life easier for new clients.

Stop recommending an encryption algorithm that is considered deprecated.

## What

* Be more permissive in the JWS alg header and the JWE enc header. The permissable alternatives are all supported by the DCS and are at least as secure as the one we originally recommended.
* Change the JWE alg header to mandate the more secure version of the header. `RSA-OAEP` is still valid, but it uses SHA-1, so is widely considered deprecated. We should therefore discourage new clients from using it.
* Update a link to a draft RFC to the final approved RFC.
* As per RFC 7518, JOSE stands for 'JSON Object Signing and Encryption'. Update accordingly.

### DIFF
--- a/source/sign-and-encrypt-a-DCS-payload.html.md.erb
+++ b/source/sign-and-encrypt-a-DCS-payload.html.md.erb
@@ -97,7 +97,7 @@ The JWE headers contain information about how the payload is encrypted.
 
 You must set the following headers:
 
-* `enc` must be set to `A128CBC-HS256` or any other [value permitted by RFC 7518](https://tools.ietf.org/html/rfc7518#section-5.1)
+* `enc` must be set to `A128CBC-HS256` or any other [value permitted by RFC 7518][supported-JWE-encodings]
 * `alg` must be set to `RSA-OAEP-256`
 * `typ` must be set to `JWE`
 * `x5t` contains the SHA1 thumbprint of the DCS encryption certificate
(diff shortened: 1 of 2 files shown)


mrwilson:
Minor nit: We have multiple references to the new RFC url, can we pull it into `_links.erb`?

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -97,8 +97,8 @@ The JWE headers contain information about how the payload is encrypted.
 
 You must set the following headers:
 
-* `enc` must be set to `A128CBC-HS256`
-* `alg` must be set to `RSA-OAEP`
+* `enc` must be set to `A128CBC-HS256` or any other [value permitted by RFC 7518][supported-JWE-encodings]
```
PippaClarkGDS, line 100: when would a user use 'any other value' rather than A128CBC-HS256?

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -110,20 +110,20 @@ Your header should look like this:
...
 ```
 ### Using encryption algorithms
 
-The encrypted component of the JavaScript Object Signing and Encryption (JOSE) message must use:
+The encrypted component of the JSON Object Signing and Encryption (JOSE) message must use:
 
-* `RSA-OAEP` as the algorithm
-* `A128CBC-HS256` as the encryption method
+* `RSA-OAEP-256` as the algorithm
+* any encryption method [permitted by RFC 7518][supported-JWE-encodings], e.g. `A128CBC-HS256`
```
PippaClarkGDS, line 124:
```suggestion
* any encryption method [permitted by RFC 7518][supported-JWE-encodings], for example `A128CBC-HS256`
```

source/sign-and-encrypt-a-DCS-payload.html.md.erb
bjgill: :shrug: possibly either because they wanted to use a more secure algorithm, or because their JOSE library has a different default?

I suspect this makes more sense re-worded to match the equivalent below

```suggestion
* `enc` must be set to a value [permitted by RFC 7518][supported-JWE-encodings], for example `A128CBC-HS256`
```

PR #171, opened 2020-05-19
H3WCW3iA: provide more detail on rate limits

bjgill:
## Why

Since the rate limit text was first written, we have agreed our initial rate limit. Since we now know how long the window will be, we can tell clients to help them prepare.

We cannot give a precise number at this stage. Commercial confidentiality means we will be unable to until all possible pilot participants have decided one way or the other. In the interim, use an indicative number that is within an order of magnitude of the expected final number (i.e. in the range 1 to 100).

## What

Provide more guidance on the expected rate limit for pilot participants.

### DIFF
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -210,7 +210,7 @@ As a pilot participant, you will be subject to the following rate limits. Both u
 1. An overall rate limit of 50 requests every 10 seconds. This is shared between all pilot participants.
 1. An individual rate limit of (300 / number of participants) requests every 60 seconds. For instance, if 10 participants join the pilot, you would have an individual rate limit of 30 requests every 60 seconds.
 
-We may change these rate limits in future to better meet your needs.
+These rate limits may change over the course of the pilot.
 
 ### Tracking your rate limit usage
 
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -203,12 +203,10 @@ A list of strings which describes the error. This field is present if there was 
 
 ## Understand rate limits in the DCS
 
-The DCS can handle up to 50 requests in total every 10 seconds.
+As a pilot participant, you'll be subject to both:
 
-As a pilot participant, you will be subject to the following rate limits. Both use a sliding window rate limit algorithm.
-
-1. An overall rate limit of 50 requests every 10 seconds. This is shared between all pilot participants.
-1. An individual rate limit of (300 / number of participants) requests every 60 seconds. For instance, if 10 participants join the pilot, you would have an individual rate limit of 30 requests every 60 seconds.
+* an overall rate limit of 50 requests in any 10 second period - this is shared between all pilot participants
+* an individual rate limit of (300 / number of participants) requests in any 60 second period - for example if 10 participants join the pilot, you would have an individual rate limit of 30 requests in any 60 second period
 
 These rate limits may change over the course of the pilot.
 
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -203,9 +203,14 @@ A list of strings which describes the error. This field is present if there was 
 
 ## Understand rate limits in the DCS
 
-DCS handles up to 50 requests in total every 10 seconds.
+The DCS can handle up to 50 requests in total every 10 seconds.
 
-This rate limit is divided fairly between the organisations connected to DCS for the pilot.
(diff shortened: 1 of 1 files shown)


bjgill:
@benjamin-mortimer - raised as suggested in our Slack discussion.

@PippaClarkGDS - would appreciate a review to see whether the wording here is understandable.

source/api-reference.html.md.erb
```diff
@@ -203,9 +203,14 @@ A list of strings which describes the error. This field is present if there was
...
-DCS handles up to 50 requests in total every 10 seconds.
+The DCS can handle up to 50 requests in total every 10 seconds.
 
-This rate limit is divided fairly between the organisations connected to DCS for the pilot.
+As a pilot participant, you will be subject to the following rate limits. Both use a sliding window rate limit algorithm.
+
+1. An overall rate limit of 50 requests every 10 seconds. This is shared between all pilot participants.
+1. An individual rate limit of (300 / number of participants) requests every 60 seconds. For instance, if 10 participants join the pilot, you would have an individual rate limit of 30 requests every 60 seconds.
+
+We may change these rate limits in future to better meet your needs.
```
benjamin-mortimer, line 213:
```suggestion
These rate limits may change over the course of the pilot.
```

Suggest tweaking this as we might have other drivers for changing the rate limit.

(Also I have what turns out to be a personal style preference for avoiding "we", which I'm sure used to be in the GOV.UK style guide...)

benjamin-mortimer:
@bjgill all looks good, but I remembered I'm not actually a code owner!
@mrwilson could you oblige again?

source/api-reference.html.md.erb
```diff
@@ -203,9 +203,14 @@ A list of strings which describes the error. This field is present if there was
 
 ## Understand rate limits in the DCS
 
-DCS handles up to 50 requests in total every 10 seconds.
+The DCS can handle up to 50 requests in total every 10 seconds.
```
PippaClarkGDS, line 206:
```suggestion
The DCS can handle up to 50 requests every 10 seconds.
```

source/api-reference.html.md.erb
```diff
@@ -203,9 +203,14 @@ A list of strings which describes the error. This field is present if there was
...
 
-This rate limit is divided fairly between the organisations connected to DCS for the pilot.
+As a pilot participant, you will be subject to the following rate limits. Both use a sliding window rate limit algorithm.
```
PippaClarkGDS, line 208:
```suggestion
As a pilot participant, you'll be subject to the following rate limits. Both use a sliding window rate limit algorithm.
```

PippaClarkGDS: will the user know what a 'sliding window rate algorithm' is? Is there a more plain English alternative we can use?

source/api-reference.html.md.erb
```diff
@@ -203,9 +203,14 @@ A list of strings which describes the error. This field is present if there was
...
+
+1. An overall rate limit of 50 requests every 10 seconds. This is shared between all pilot participants.
+1. An individual rate limit of (300 / number of participants) requests every 60 seconds. For instance, if 10 participants join the pilot, you would have an individual rate limit of 30 requests every 60 seconds.
+
```
PippaClarkGDS, line 212:
```suggestion

* an overall rate limit of 50 requests every 10 seconds - this is shared between all pilot participants
* an individual rate limit of (300 / number of participants) requests every 60 seconds - for example if 10 participants join the pilot, you would have an individual rate limit of 30 requests every 60 seconds

```

PippaClarkGDS: does the user use both of these rate limits or is it an either/or? Confused how these two points interact with each other?

source/api-reference.html.md.erb
bjgill: "sliding window rate limit" is a term of art - I don't think there's an alternative.

source/api-reference.html.md.erb
bjgill: Both apply simultaneously.

bjgill: I wonder if we could make this less confusing by rewording slightly:

"""
As a pilot participant, you'll be subject to both:

* an overall ...
* an individual ...

The DCS uses a a sliding window rate limit algorithm. These rate limits may change over the course of the pilot.
"""

bjgill: Had a chat via Slack and agreed new wording

PR #174, opened 2020-05-20
Explain you must pass the 'review' stage before you get prod access

benjamin-mortimer:
This is a condition we have not mentioned in the docs. We'll be seeing some integration connections soon so worth being clear about this here.

### DIFF
--- a/source/integrating-with-the-dcs.html.md.erb
+++ b/source/integrating-with-the-dcs.html.md.erb
@@ -1,7 +1,7 @@
 ---
 title: How the DCS works
 weight: 2
-last_reviewed_on: 2020-05-04
+last_reviewed_on: 2020-05-20
 review_in: 6 weeks
 ---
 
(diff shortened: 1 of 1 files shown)


source/integrating-with-the-dcs.html.md.erb
```diff
@@ -34,7 +34,10 @@ To build a working DCS client, you'll need to:
 
 ## Getting access to production
 
-You’ll need to successfully complete a full journey in the test environment before you can get access to the production environment.
+Before you can access production, you'll need to:
+
++ successfully complete a full journey in the test environment
```
mrwilson, line 39: Do we refer to it as a 'test environment' or 'integration environment' elsewhere in the docs? We should be consistent either way.

source/integrating-with-the-dcs.html.md.erb
```diff
@@ -34,7 +34,10 @@ To build a working DCS client, you'll need to:
...
++ show your organisation meet the [DCS pilot requirements](https://www.gov.uk/guidance/requirements-for-joining-the-document-checking-service-pilot) - the DCS team will contact you about this separately
```
mrwilson:
```suggestion
+ show your organisation meets the [DCS pilot requirements](https://www.gov.uk/guidance/requirements-for-joining-the-document-checking-service-pilot) - the DCS team will contact you about this separately.
```

source/integrating-with-the-dcs.html.md.erb
benjamin-mortimer: Yeah, good shout - turns out this is the only reference to it as 'test' so have changed it to 'integration'.

source/integrating-with-the-dcs.html.md.erb
benjamin-mortimer: Have fixed the verb but the missing full stop is GOV.UK style

PR #175, opened 2020-05-20
Remind participants to handle keys securely

benjamin-mortimer:
It's important!

### DIFF
--- a/source/generate-keys-and-request-certificates.html.md.erb
+++ b/source/generate-keys-and-request-certificates.html.md.erb
@@ -28,12 +28,12 @@ This documentation gives examples of how you’d follow this process using OpenS
 
 You must make sure you:
 
-+ store your key material securely, in a safe place
++ store your keys in a safe place
 + limit access to your key material
 
-If you suspect your key material has been compromised, [tell the DCS support team immediately](/support/#security) and stop using the compromised material.
+If you suspect your keys have been compromised, [tell the DCS support team immediately](/support/#security) and stop using the compromised keys.
 
-You must handle your integration key material as securely as your production key material. The DCS integration environment is a live service.
+You must handle your integration keys as securely as your production keys. The DCS integration environment is a live service.
 
 ## Generate a private RSA key
 
--- a/source/generate-keys-and-request-certificates.html.md.erb
+++ b/source/generate-keys-and-request-certificates.html.md.erb
@@ -31,7 +31,7 @@ You must make sure you:
 + store your key material securely, in a safe place
 + limit access to your key material
 
-If you suspect your key material has been compromised, [tell the DCS support team immediately](https://dcs-pilot-docs.cloudapps.digital/support/#security) and stop using the compromised material.
+If you suspect your key material has been compromised, [tell the DCS support team immediately](/support/#security) and stop using the compromised material.
 
 You must handle your integration key material as securely as your production key material. The DCS integration environment is a live service.
 
--- a/source/generate-keys-and-request-certificates.html.md.erb
+++ b/source/generate-keys-and-request-certificates.html.md.erb
@@ -31,7 +31,7 @@ You must make sure you:
 + store your key material securely, in a safe place
 + limit access to your key material
 
-If you suspect your key material has been compromised, tell GDS immediately and stop using the compromised material.
+If you suspect your key material has been compromised, [tell GDS immediately](https://dcs-pilot-docs.cloudapps.digital/support/#security) and stop using the compromised material.
 
 You must handle your integration key material as securely as your production key material. The DCS integration environment is a live service.
 
--- a/source/generate-keys-and-request-certificates.html.md.erb
+++ b/source/generate-keys-and-request-certificates.html.md.erb
@@ -31,7 +31,7 @@ You must make sure you:
 + store your key material securely, in a safe place
 + limit access to your key material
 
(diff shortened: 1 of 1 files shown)


source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -24,6 +24,17 @@ The GDS CA must sign and issue the certificates.
...
 This documentation gives examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer.
 
+## Keep your keys secure
+
+You must make sure you:
+
++ store your key material securely, in a safe place
++ limit access to your key material
+
+If you suspect your key material has been compromised, tell GDS immediately and stop using the compromised material.
```
bjgill, line 34: > tell GDS immediately

Are we able to say how they should do this?

benjamin-mortimer: I've added a link to the contact page there - for security issues, we ask them to use the support email, as for other things.

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -24,6 +24,17 @@ The GDS CA must sign and issue the certificates.
 
 This documentation gives examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer.
 
+## Keep your keys secure
+
+You must make sure you:
+
++ store your key material securely, in a safe place
```
PippaClarkGDS, line 31: a couple of questions: what is key material? Can we just say 'keys'? 

Also is securely and in a safe place a tautology (a safe place is usually secure)? If so, let's remove 'a safe place' and if there's an example we can give, then provide one

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -24,6 +24,17 @@ The GDS CA must sign and issue the certificates.
...
++ limit access to your key material
+
+If you suspect your key material has been compromised, [tell GDS immediately](https://dcs-pilot-docs.cloudapps.digital/support/#security) and stop using the compromised material.
```
PippaClarkGDS, line 34:
```suggestion
If you suspect your key material has been compromised, [tell the DCS support team immediately](https://dcs-pilot-docs.cloudapps.digital/support/#security) and stop using the compromised material.
```

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -24,6 +24,17 @@ The GDS CA must sign and issue the certificates.
...
+
+You must handle your integration key material as securely as your production key material. The DCS integration environment is a live service.
```
PippaClarkGDS, line 36: again - what does material mean here?

source/generate-keys-and-request-certificates.html.md.erb
bjgill: Would recommend using a relative link. That way the link will continue to work if the base URL ever changes.

```suggestion
If you suspect your key material has been compromised, [tell GDS immediately](/support/#security) and stop using the compromised material.
```

source/generate-keys-and-request-certificates.html.md.erb
benjamin-mortimer: Makes sense, have changed

source/generate-keys-and-request-certificates.html.md.erb
benjamin-mortimer: Have changed to just "keys"

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -24,6 +24,17 @@ The GDS CA must sign and issue the certificates.
 
 This documentation gives examples of how you’d follow this process using OpenSSL. You can use other methods if you prefer.
 
+## Keep your keys secure
+
+You must make sure you:
+
++ store your keys in a safe place
```
PippaClarkGDS, line 31: is it worth mentioning what a 'safe place' would look like as it could be interpreted differently?

source/generate-keys-and-request-certificates.html.md.erb
```diff
@@ -24,6 +24,17 @@ The GDS CA must sign and issue the certificates.
...
++ limit access to your key material
+
+If you suspect your keys have been compromised, [tell the DCS support team immediately](/support/#security) and stop using the compromised keys.
```
PippaClarkGDS, line 34:
```suggestion
If you suspect your keys have been compromised: 
* [tell the DCS support team immediately](/support/#security)
* stop using the compromised keys
```

benjamin-mortimer: I'd rather keep it as a one line sentence to be clear you have to do both things - having it as a one line sentence means you can use 'and'.

source/generate-keys-and-request-certificates.html.md.erb
benjamin-mortimer: I've changed it to "securely", because there are a variety of places you could keep your keys and it doesn't matter too much as long as they're secure.

There are also a variety of ways of keeping them secure, but we don't have to explain those here - it's outside the scope of "how to connect to DCS".

source/generate-keys-and-request-certificates.html.md.erb
PippaClarkGDS: I get you, but think the points will get blurred if blended into one line - how about using a numbered list instead of bullets to make it clear users must do both steps?

benjamin-mortimer:
I've made it a bulleted list - let's revisit if people actually find it confusing.

PR #176, opened 2020-05-29
BAU: Scheduled review requested by the spaniel

chrisclayson:
## Why

Review prompted by Daniel the Manual Spaniel.

## What

Update review dates.
Include entry on `status-response-codes` page for `400` (Bad request), this should probably be fleshed out in a story.

Removed references to `503` planned outage and planned outage functionality in status responses as this is a deprecated feature.

### DIFF
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -1,8 +1,8 @@
 ---
 title: API reference
 weight: 6
-last_reviewed_on: 2020-05-06
-review_in: 3 weeks
+last_reviewed_on: 2020-05-29
+review_in: 6 weeks
 ---
 
 # API reference
@@ -43,10 +43,11 @@ If the DCS is available, the JSON response body will be similar to:
 }
 ```
 
+Please note the `scheduledOutages` field is deprecated and will always contain and empty array.
+
 If the DCS is unavailable, the `message` field in the response describes the cause. The cause for unavailability can be:
 
 * `Limit exceeded` - you have exceeded your allocated number of API calls
-* `Planned outage in progress finishing <TIMESTAMP>` - the DCS or HM Passport Office (HMPO) are in a period of scheduled downtime, ending at `<TIMESTAMP>`
 * `Unplanned outage` - the DCS or HMPO are experiencing a period of unscheduled downtime
 
 For example, during a planned outage, the response is similar to:
@@ -54,15 +55,7 @@ For example, during a planned outage, the response is similar to:
 ```json
 {
   "available": false,
-  "message": "Planned outage in progress finishing 2019-09-12T00:00:00.000Z",
-  "scheduledOutages":
-    [
-      {
-      	"start":"2019-09-05T00:00:00.000Z",
-      	"end":"2019-09-12T00:00:00.000Z",
-      	"message":"Maintenance work is underway."
-      }
-    ]
+  "message": "Unplanned outage",
 }
 ```
 
@@ -71,18 +64,6 @@ Boolean field indicating if the DCS is available and serving requests.
 
 #### message
 A string explaining why the DCS is unavailable. For example, if there is an ongoing planned outage, the field will contain detail about the duration of the outage.
-
-#### scheduledOutages
-A list of upcoming scheduled outages. The list is empty if there are no outages.
-
-#### scheduledOutages.start
-String in ISO 8601 format indicating when an outage is scheduled to start.
-
-#### scheduledOutages.end
-String in ISO 8601 format indicating when an outage is scheduled to end.
-
-#### scheduledOutages.message
-String indicating the reason for the scheduled outage.
 
 ## Check passport validity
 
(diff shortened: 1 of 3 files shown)


source/api-reference.html.md.erb
```diff
@@ -43,26 +43,19 @@ If the DCS is available, the JSON response body will be similar to:
...
 
+Please note the `scheduledOutages` field is deprecated and will always contain and empty array.
+
 If the DCS is unavailable, the `message` field in the response describes the cause. The cause for unavailability can be:
 
 * `Limit exceeded` - you have exceeded your allocated number of API calls
-* `Planned outage in progress finishing <TIMESTAMP>` - the DCS or HM Passport Office (HMPO) are in a period of scheduled downtime, ending at `<TIMESTAMP>`
 * `Unplanned outage` - the DCS or HMPO are experiencing a period of unscheduled downtime
 
 For example, during a planned outage, the response is similar to:
```
bjgill, line 53: Do we need to update this as well?

source/status-response-codes.html.md.erb
```diff
@@ -12,7 +12,8 @@ The Document Checking Service (DCS) uses standard [HTTP response code][HTTP-Stat
 | HTTP response code | Description                            |
 | ------------------ | -------------------------------------- |
 | 200                | The DCS handled your request successfully. |
-| 403                | You have used up [your quota of DCS checks][quotas] and cannot make any more DCS checks.
+| 400                | Your JOSE payload is incorrectly formed, the error message will tell you what is wrong |
```
bjgill, line 15: This is one of the reasons you might get a 400, but there are many more. Can we be less specific?

source/api-reference.html.md.erb
```diff
@@ -140,10 +121,7 @@ The passport expiration date in YYYY-MM-DD format. Must be in the future or no m
 
 A successful request is one that completes without error, regardless of whether the passport is valid or not.
 
-The DCS returns a:
-
-* `200` [HTTP response code][status-response-codes] for successful requests
-* `503` [HTTP response code][status-response-codes] if HMPO is having a planned outage
+The DCS returns a `200` [HTTP response code][status-response-codes] for successful requests, non-successful responses will return [as documented here][status-response-codes].
```
bjgill, line 124: Possibly worth talking to Pippa (or someone) about the wording here?

source/api-reference.html.md.erb
```diff
@@ -43,26 +43,19 @@ If the DCS is available, the JSON response body will be similar to:
 }
 ```
 
+Please note the `scheduledOutages` field is deprecated and will always contain and empty array.
+
 If the DCS is unavailable, the `message` field in the response describes the cause. The cause for unavailability can be:
 
 * `Limit exceeded` - you have exceeded your allocated number of API calls
-* `Planned outage in progress finishing <TIMESTAMP>` - the DCS or HM Passport Office (HMPO) are in a period of scheduled downtime, ending at `<TIMESTAMP>`
```
bjgill, line 49: Do we need to document validation failures? No need to do that in this PR, though...

source/api-reference.html.md.erb
PippaClarkGDS:
```suggestion
The DCS returns a `200` [HTTP response code][status-response-codes] for successful requests. 

If the requests are not successful, the DCS will return a [HTTP response code][status-response-codes] to indicate why the request was not successful.
```

PippaClarkGDS: My stab at making this more clear - what do you think @chrisclayson ?

source/api-reference.html.md.erb
chrisclayson: Well spotted!

source/status-response-codes.html.md.erb
chrisclayson: Agree, done!

source/api-reference.html.md.erb
chrisclayson: Sounds like a story to me, will raise one (if it doesn't exist).

source/api-reference.html.md.erb
chrisclayson: Sounds good to me. I've commited your suggestion! 👍

source/api-reference.html.md.erb
```diff
@@ -43,26 +43,19 @@ If the DCS is available, the JSON response body will be similar to:
 }
 ```
 
+Please note the `scheduledOutages` field is deprecated and will always contain and empty array.
```
PippaClarkGDS, line 46:
```suggestion
Please note the `scheduledOutages` field is deprecated and will always contain an empty array.
```

PippaClarkGDS: will a user know and understand what 'deprecated' and 'an empty array' is and how it affects them?

source/status-response-codes.html.md.erb
```diff
@@ -12,7 +12,8 @@ The Document Checking Service (DCS) uses standard [HTTP response code][HTTP-Stat
 | HTTP response code | Description                            |
 | ------------------ | -------------------------------------- |
 | 200                | The DCS handled your request successfully. |
-| 403                | You have used up [your quota of DCS checks][quotas] and cannot make any more DCS checks.
+| 400                | There is a problem with you request, the error message will tell you what is wrong. |
```
PippaClarkGDS, line 15:
```suggestion
| 400                | There is a problem with your request - the error message will tell you what is wrong and how to fix it. |
```

PippaClarkGDS: this feels a bit vague as we then elaborate on the 400 response codes just below?

source/api-reference.html.md.erb
chrisclayson: I would hope so, we could say something like "Please note the `scheduledOutages` field is an old feature that is no longer used, it will always be empty (as shown). You should ignore this field."

PR #182, opened 2020-06-10
Api reference

PippaClarkGDS:
## Why

API reference needed some user-centred design and updating

## What

Paired with Alex to write this content and update jargon to make the API reference more user-friendly. 

** need to flag: the column width is squashed on the tables. Couldn't find a way to resolve this - any tips welcome**

### DIFF
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -1,38 +1,36 @@
 ---
 title: API reference
 weight: 6
-last_reviewed_on: 2020-05-29
+last_reviewed_on: 2020-06-10
 review_in: 6 weeks
 ---
 
 # API reference
 
-The Document Checking Service (DCS) API is based on [REST principles][REST]. It returns data in [JSON format][JSON], and uses standard [HTTP status response codes][status-response-codes].
+The Document Checking Service (DCS) API:
+
+* is based on [REST principles](https://restfulapi.net/)
+* returns data in [JSON format](https://www.rfc-editor.org/info/rfc8259)
+* uses standard [HTTP status response codes](/status-response-codes/#status-response-codes)
 
 This section describes the calls you can make to the DCS API, along with example requests and possible responses.
 
 Using the DCS API you can:
 
-* [see the operational status of the DCS][check-status]
-* [check passport validity][check-passport]
+* [check the DCS status](/api-reference/#check-the-dcs-status)
+* [check whether a passport is valid](/api-reference/#check-whether-a-passport-is-valid)
 
 ## Check the DCS status
 
-You can check the status of the DCS by making a HTTP GET request to the service endpoint.
+You can check the status of the DCS which helps you build a fallback plan if the DCS is unavailable.
 
-We recommend you use this endpoint to build appropriate levels of reliability into your service.
+### Make a status request
 
-### The DCS integration endpoint
-
-The DCS integration endpoint is <https://dcs-integration.ida.digital.cabinet-office.gov.uk/>.
-
-### Request
-
-`GET: /checks/passport`
+`GET /checks/passport`
 
 The request body for a DCS status check is empty and therefore does not require the [signature and encryption wrapper](/sign-and-encrypt-a-DCS-payload/#sign-and-encrypt-a-dcs-payload).
 
-### Response
+### Interpret the status response
 
 If the DCS is available, the JSON response body will be similar to:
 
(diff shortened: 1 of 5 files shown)


PippaClarkGDS:
** need to flag: the column width is squashed on the tables. Couldn't find a way to resolve this - any tips welcome**

source/api-reference.html.md.erb
```diff
@@ -43,89 +41,65 @@ If the DCS is available, the JSON response body will be similar to:
...
-* `Unplanned outage` - the DCS or HMPO are experiencing a period of unscheduled downtime
+* `Limit exceeded` - you have exceeded your [rate limit](/api-reference/#understand-rate-limits-in-the-dcs)
+* `Unplanned outage` - the DCS or Her Majesty’s Passport Office (HMPO) is experiencing a period of unscheduled downtime
 
-For example, during a unplanned outage, the response will be:
+| Response field                              | Type    | Description                                                                                                                                              |
+|------------------------------------|---------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `available`                | boolean | Indicates if the DCS is available and serving requests.                                                                                                  |
+| `scheduledOutages`         | array   | A list of upcoming scheduled outages, which is empty if there are no outages scheduled.                                                                  |
```
bjgill, line 52: Are scheduled outages still a thing? I thought this field was always going to be empty?

source/api-reference.html.md.erb
```diff
@@ -43,89 +41,65 @@ If the DCS is available, the JSON response body will be similar to:
...
+|--------------------------------------------|--------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `correlationId`  | string | Contains an [RFC 4122](https://tools.ietf.org/html/rfc4122)-compliant Universally Unique Identifier (UUID) which ties together multiple requests in the same session.                                                                                        |
+| `requestId`      | string | Contains an [RFC 4122](https://tools.ietf.org/html/rfc4122)-compliant UUID which identifies a single request within a session.                                                                                                                               |
+| `timestamp`      | string | When your client makes the request. Must be in [ISO 8601](https://www.w3.org/TR/NOTE-datetime), for example 2020-01-01T12:30:00.000Z.                                                                                                                        |
```
bjgill, line 74: Possibly worth adding that the timestamp must be no more than 5 minutes in the past/future?

source/api-reference.html.md.erb
```diff
@@ -43,89 +41,65 @@ If the DCS is available, the JSON response body will be similar to:
...
+| `passportNumber` | number | The number on the passport. This is between 1 and 899999999.                                                                                                                                                                                                 |
```
benjamin-mortimer, line 75: Isn't it actually a string?

source/api-reference.html.md.erb
```diff
@@ -160,64 +134,49 @@ Example response body for a successful response containing an error:
...
-
-#### correlationId
-String containing an [RFC 4122] compliant UUID which ties together multiple requests in the same session.
-
-#### requestId
-String containing an [RFC 4122] compliant UUID which identifies a single request within a session. This is the request ID from the original request.
+| Response field          | Type    | Description                                                                                                                              |
+|-----------------|---------|------------------------------------------------------------------------------------------------------------------------------------------|
+| `correlationId` | string  | Contains an [RFC 4122](https://tools.ietf.org/html/rfc4122)-compliant UUID which ties together multiple requests in the same session.    |
+| `requestId`     | string  | Contains an [RFC 4122](https://tools.ietf.org/html/rfc4122)-compliant UUID which identifies a single request within a session.           |
```
bjgill, line 140: Would it help to be explicit that these two IDs will be identical to those provided in the request?

source/api-reference.html.md.erb
```diff
@@ -160,64 +134,49 @@ Example response body for a successful response containing an error:
...
 ## Understand rate limits in the DCS
 
-As a pilot participant, you'll be subject to both:
+The DCS handles up to 50 requests every 10 seconds.
 
-* an overall rate limit of 50 requests in any 10 second period - this is shared between all pilot participants
-* an individual rate limit of (300 / number of participants) requests in any 60 second period - for example if 10 participants join the pilot, you would have an individual rate limit of 30 requests in any 60 second period
-
-These rate limits may change over the course of the pilot.
+This rate limit is divided between the organisations connected to the DCS for the pilot.
```
bjgill, line 150: Why are we giving less detail on the rate limits?

source/api-reference.html.md.erb
```diff
@@ -1,38 +1,36 @@
...
+last_reviewed_on: 2020-06-10
 review_in: 6 weeks
 ---
 
 # API reference
 
-The Document Checking Service (DCS) API is based on [REST principles][REST]. It returns data in [JSON format][JSON], and uses standard [HTTP status response codes][status-response-codes].
+The Document Checking Service (DCS) API:
+
+* is based on [REST principles](https://restfulapi.net/)
```
bjgill, line 12: Why are we putting these links inline (as opposed to using the link partial i.e. `[REST]`)?

source/api-reference.html.md.erb
PippaClarkGDS: flagging to @mrwilson to confirm

source/api-reference.html.md.erb
PippaClarkGDS: flagging to @mrwilson to confirm

source/api-reference.html.md.erb
PippaClarkGDS: flagging to @mrwilson - I think we decided to omit the exact limits of when a timestamp is/isn't valid for security reasons?

source/api-reference.html.md.erb
PippaClarkGDS: cheers Benjamin - what do you mean and I'll include if needed?

source/api-reference.html.md.erb
PippaClarkGDS: flagging to @mrwilson to confirm - happy to include more detail if it's useful to the users

source/api-reference.html.md.erb
PippaClarkGDS: I can just include 'REST' as the link - thanks for flagging

PippaClarkGDS: will do this for the JSON just below too

source/api-reference.html.md.erb
bjgill: Can confirm that this is a string:

```
private final String passportNumber;
```

source/api-reference.html.md.erb
bjgill: Possibly something like this?

```suggestion
| `requestId`     | string  | Contains the same `requestId` provided in the request, which is a [RFC 4122]-compliant UUID. |
```

source/api-reference.html.md.erb
mrwilson: Technically "yes" insofar as the request will always contain a `scheduledOutages` field - it's just that this field will never be non-empty, as we've hard-coded it to be empty and no longer support this feature.

I'm wondering about the feasibility of dropping this field entirely from pilot clients, as we'll have access to that information when we return the status response.

source/api-reference.html.md.erb
mrwilson: I like this wording.

source/api-reference.html.md.erb
mrwilson: For the time being, let's keep it in?

source/api-reference.html.md.erb
mrwilson: We thought about this during the first draft.

I was unable to find any instances of this validation error being triggered in the last 3 months within the audit database.

I would assert that the meaning is clear. Let's revisit this in the future?

PR #192, opened 2020-07-01
added gdoc content about DCS responses

PippaClarkGDS:
## Why

Need documentation about what responses mean on the DCS

## What

Documents what the DCS responses mean.

### DIFF
--- a/source/integrating-with-the-dcs.html.md.erb
+++ b/source/integrating-with-the-dcs.html.md.erb
@@ -64,7 +64,7 @@ It’s possible for a valid passport to return a `valid: false` response. Check 
 
 ### Receiving an `error: true` response
 
-An `error: true` response means the HMPO database is unavailable. You should retry your request.
+An `error: true` response means the HMPO database is unavailable. You should wait a few seconds and then try again.
 
 ## Getting access to production
 
--- a/source/integrating-with-the-dcs.html.md.erb
+++ b/source/integrating-with-the-dcs.html.md.erb
@@ -1,8 +1,8 @@
 ---
 title: How the DCS works
 weight: 2
-last_reviewed_on: 2020-05-20
-review_in: 6 weeks
+last_reviewed_on: 2020-06-20
+review_in: 8 weeks
 ---
 
 # How the DCS works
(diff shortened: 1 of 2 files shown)


source/integrating-with-the-dcs.html.md.erb
```diff
@@ -36,6 +36,36 @@ Currently, there are no supported client libraries for the DCS.
...
+* `valid: false`
+* `error: true`
+
+### Receiving a `valid: true` response
+
+A `valid: true` response means:
+
+* all the details on the passport, for example, first names or date of birth, exactly match the details held in Her Majesty’s Passport Office (HMPO)
+* the passport is not marked as lost, stolen or cancelled in the HMPO database
+* the passport has not expired, or expired no more than 18 months ago
```
bjgill, line 53: There are a couple of edge cases not covered by this list? Is this an intentional omission?

See the last two bullet points of a page in our internal docs: https://dcs-service-manual.cloudapps.digital/about-dcs/architecture/data-source-messages/#success

source/integrating-with-the-dcs.html.md.erb
```diff
@@ -36,6 +36,36 @@ Currently, there are no supported client libraries for the DCS.
...
+
+* one or more of the passport details, for example first names or date of birth, does not exactly match the details in the HMPO database
+* the passport is marked as lost, stolen or cancelled in the HMPO database
+* HMPO has not yet sent the passport to the holder
+
+It’s possible for a valid passport to return a `valid: false` response. Check how to [reduce errors when using data from a passport chip](/reading-data-from-passport-nfc-chip/#reduce-errors-when-using-data-from-a-passport-chip).
+
+### Receiving an `error: true` response
+
+An `error: true` response means the HMPO database is unavailable. You should retry your request.
```
bjgill, line 67: > You should retry your request.

Could we give some guidance here? If they retry immediately and automatically on failure, they are likely to start hitting rate limits if HMPO is down.

source/integrating-with-the-dcs.html.md.erb
PippaClarkGDS: cc @benjamin-mortimer as he was the SME for this info

source/integrating-with-the-dcs.html.md.erb
PippaClarkGDS: cc @benjamin-mortimer as he was the SME for this info - I know we can't put exact time limits on but is there anything we could add here?

bjgill: Maybe something like "You should wait a few seconds and then try again"?

PippaClarkGDS: happy with this from a content perspective ^^ @benjamin-mortimer checking this works ok?

PippaClarkGDS:
> Looks useful - just a couple of questions about the details of the content.
> 
> Could we make it easier for users to find the information they need about the DCS response? It looks as if we've got information about DCS responses spread out quite widely - it's already somewhat present in https://dcs-pilot-docs.cloudapps.digital/access-the-data-in-a-DCS-response/#review-and-use-the-human-readable-response and https://dcs-pilot-docs.cloudapps.digital/api-reference/#interpret-the-passport-validity-response. Is there a risk that users look in the API reference, and miss this useful information?

>> good point, Benjamin. I'll add in some links on those pages and anywhere else that seems relevant to point users in this direction and make them aware this is documented.

source/integrating-with-the-dcs.html.md.erb
benjamin-mortimer: Yes, the omission is intentional. Might add in later, however.

source/integrating-with-the-dcs.html.md.erb
benjamin-mortimer: Yeah - @bjgill  knows more about this than I do.

benjamin-mortimer:
Yes, we should come back to this and make sure we have one thing per need. However, there's some important info in this PR that we shouldn't delay. I'll add a card to the backlog to rethink the "DCS responses" user journey.

source/integrating-with-the-dcs.html.md.erb
PippaClarkGDS: great stuff - will amend

PippaClarkGDS:
I can't add in links throughout the documentation yet as the local host isn't working. Going to push this main information live and make a card to add in links pointing to this so it can be actioned later.

PR #203, opened 2020-08-24
7LG7Kop7: Remove SHA1 thumbprint as a necessary JWS header

mrwilson:
## Why

These docs claim that we mandate both SHA1 and SHA256 thumbprint headers in the JWS payload layers, while this isn't strictly true.

## What

This PR removes references to SHA1 from the doc as SHA256 is the preferred thumbprint.


PippaClarkGDS (approved):
Looks good

PippaClarkGDS (approved):

PR #208, opened 2020-09-08
Explain you must sign a contract before enrolling

benjamin-mortimer:
Pilot participants need to sign a contract before they can begin the CA enrolment process. We should point this out in our documentation.

### DIFF
--- a/source/enrol-with-the-DCS-certificate-authority.html.md.erb
+++ b/source/enrol-with-the-DCS-certificate-authority.html.md.erb
@@ -7,7 +7,9 @@ review_in: 2 months
 
 # Enrol with the DCS certificate authority
 
-Before you can [raise certificate signing requests (CSRs)](/generate-keys-and-request-certificates/#create-a-certificate-signing-request-csr), you’ll need to enrol with the DCS certificate authority (CA).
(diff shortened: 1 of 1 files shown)


source/enrol-with-the-DCS-certificate-authority.html.md.erb
```diff
@@ -7,7 +7,9 @@ review_in: 2 months
 
 # Enrol with the DCS certificate authority
 
-Before you can [raise certificate signing requests (CSRs)](/generate-keys-and-request-certificates/#create-a-certificate-signing-request-csr), you’ll need to enrol with the DCS certificate authority (CA).
+Before you can [raise certificate signing requests (CSRs)](/generate-keys-and-request-certificates/#create-a-certificate-signing-request-csr), you need to enrol with the DCS certificate authority (CA).
+
+You need to sign a contract for the DCS pilot before you can enrol with the CA.
```
PippaClarkGDS, line 12: might be worth having a call to action for this eg "<Contact DCS Support> to get your contract." or similar

benjamin-mortimer: I've updated this to say to contact the support email - don't think there's much else we can advise people to do in the public docs.

PippaClarkGDS: nice - it works well

PippaClarkGDS (approved):

PR #214, opened 2020-09-23
fixed headings in accessibility to be compliant with regulations

PippaClarkGDS:
## Why

Need specific headings in our accessibility statement to make any non-compliances clear.

## What

Need specific headings in our accessibility statement to make any non-compliances clear.


chrisclayson (approved):
👍

PR #217, opened 2020-09-25
AfMQJza1: Add table of possible validation error messages

mrwilson:
## Why

Pilot clients will be able to diagnose errors on their side more easily if we explain what each error means.

## What

A table of all possible validation error messages for Passport Validation requests

### DIFF
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -92,6 +92,43 @@ The DCS validates each field before forwarding the evidence check to HMPO. For e
   "requestId": "not-a-valid-uuid"
 }
 ```
+
+#### Request validation error messages
+
+If any of the request fields do not pass validation, the `errorMessage` field will contain an entry for each validation error detailing the cause of the error.
+
+These are in the format `validation: <field>: <message>`, for example `validation: requestId: must be in universally unique identifier (UUID) format`.
+
+We've listed all possible validation error messages for requests to the DCS's passport checking endpoint, along with some example values that would cause each error message.
+
+| Request field       | Error message   | Example of what could cause the error message |
+|--------------------------------------------|--------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+requestId|must be in universally unique identifier (UUID) format|`1234`
+||must not be empty|`<blank>`
+correlationId|must be in universally unique identifier (UUID) format|`1234`
+||must not be empty|`<blank>`
+timestamp|must be in ISO 8601 format YYYY-MM-DD'T'hh:mm:ss.SSSZ. For example, 2020-01-01T12:30:00.000Z|`01-01-1990 12:00:00`
+||must not be empty|`<blank>`
+||is not recent enough|`1990-01-01T12:00:00.000Z`
+passportNumber|must be a numeric string between 1 and 899999999|`999999999`
+||must be a numeric string between 1 and 899999999|`abcde`
+||must not be empty|`<blank>`
+surname|length must be between 1 and 30 characters|`long...name`
+||must only contain the characters: a-z, A-Z, space, hyphen, full stop and apostrophe|`12345`
+||must not be empty|`<blank>`
+forenames|must not be empty|`[]`
+||length must be between 1 and 30 characters|`["long...", "...names"]`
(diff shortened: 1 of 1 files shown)


source/api-reference.html.md.erb
```diff
@@ -92,6 +92,43 @@ The DCS validates each field before forwarding the evidence check to HMPO. For e
   "requestId": "not-a-valid-uuid"
 }
 ```
+
+#### Request Validation Error Messages
```
PippaClarkGDS, line 96:
```suggestion
#### Request validation error messages
```

source/api-reference.html.md.erb
```diff
@@ -92,6 +92,43 @@ The DCS validates each field before forwarding the evidence check to HMPO. For e
...
+
+The `errorMessage` field will contain an entry for each validation error if any of request fields do not pass validation as specified in the API reference.
```
PippaClarkGDS, line 98:
```suggestion
If any of the request fields do not pass validation, the `errorMessage` field will contain an entry for each validation error detailing the cause of the error.
```

source/api-reference.html.md.erb
```diff
@@ -92,6 +92,43 @@ The DCS validates each field before forwarding the evidence check to HMPO. For e
...
+
+These are in the format `validation: <field>: <message>`, such as `validation: requestId: must be in universally unique identifier (UUID) format`
```
PippaClarkGDS, line 100:
```suggestion
These are in the format `validation: <field>: <message>`, for example `validation: requestId: must be in universally unique identifier (UUID) format`.
```

source/api-reference.html.md.erb
```diff
@@ -92,6 +92,43 @@ The DCS validates each field before forwarding the evidence check to HMPO. For e
...
+
+We've listed all possible validation error messages for requests to the DCS' passport checking endpoint, alongside some example values that would cause each error message.
```
PippaClarkGDS, line 102:
```suggestion
We've listed all possible validation error messages for requests to the DCS's passport checking endpoint, along with some example values that would cause each error message.
```

source/api-reference.html.md.erb
```diff
@@ -92,6 +92,43 @@ The DCS validates each field before forwarding the evidence check to HMPO. For e
...
+
+| Field       | Error Message   | Example Bad Input|
```
PippaClarkGDS, line 104: think there is a missing row underneath `passportNumber`?

PippaClarkGDS, line 104:
```suggestion
| Request field       | Error message   | Example of what could cause the error message |
```

PippaClarkGDS (commented):
Looks good, Alex - a few small changes. Let me know if anything needs clarifying

source/api-reference.html.md.erb
mrwilson: This is intended - there are multiple ways to trigger that particular error message: an otherwise-valid passport number beginning with `9` _or_ a non-numeric string

mrwilson:
I've taken all your feedback on board, thanks Pippa!

source/api-reference.html.md.erb
```diff
@@ -92,6 +92,43 @@ The DCS validates each field before forwarding the evidence check to HMPO. For e
...
+|--------------------------------------------|--------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+requestId|must be in universally unique identifier (UUID) format|`1234`
+||must not be empty|`<blank>`
+correlationId|must be in universally unique identifier (UUID) format|`1234`
+||must not be empty|`<blank>`
+timestamp|must be in ISO 8601 format YYYY-MM-DD'T'hh:mm:ss.SSSZ. For example, 2020-01-01T12:30:00.000Z|`01-01-1990 12:00:00`
+||must not be empty|`<blank>`
+||is not recent enough|`1990-01-01T12:00:00.000Z`
+passportNumber|must be a numeric string between 1 and 899999999|`999999999`
+|||`abcde`
```
PippaClarkGDS, line 114: For accessibility, we'd need to include a row detailing what could cause this error - I'd recommend including the line from the cell above to make it clear

PippaClarkGDS (commented):
Looks good! One small change left and then it looks good to push

PippaClarkGDS (approved):
Looks good, nice work Alex

PR #225, opened 2020-10-12
Explain how to request access to production

benjamin-mortimer:
## Why

We've been in touch with some organisations about what they need to do to get access to production. We should explain that in our documentation.

### DIFF
--- a/source/getting-access-to-production.html.md.erb
+++ b/source/getting-access-to-production.html.md.erb
@@ -9,10 +9,10 @@ review_in: 8 weeks
 
 After you have built a working client, you can request access to the DCS production environment.
 
-[Contact the DCS team](/support) to request access to production. You'll need to confirm that:
+[Contact the DCS team](/support) to request access to production. You'll need to confirm your service:
 
-+ your service is ready to use - that is, once you are connected to production someone could use it for the purpose set out in the use case you submitted as part of your application for the DCS pilot
-+ your service has successfully handled the three outcomes from the [test requests](/send-a-test-request-to-the-DCS) you sent to the DCS integration environment (`valid: true`, `valid: false` and `error: true`)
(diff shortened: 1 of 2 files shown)


benjamin-mortimer:
@PippaClarkGDS what do you think of this? 

We had some info about how to request production access but it's a bit out of date.

In updating it, I thought it also made sense to add this as its own guide as the final step of the connecting steps.

source/getting-access-to-production.html.md.erb
```diff
@@ -0,0 +1,18 @@
...
+
+# Request access to production
+
+After you have built a working client, you can request access to the DCS production environment.
+
+[Contact the DCS team](/support) to request access to production. You'll need to confirm that:
+
++ your service is ready to use - that is, once you are connected to production someone could use it for the purpose set out in the use case you submitted as part of your application for the DCS pilot
++ your service has successfully handled the three outcomes from the [test requests](/send-a-test-request-to-the-DCS) you sent to the DCS integration environment (`valid: true`, `valid: false` and `error: true`)
++ you have an auditing system set up to keep a record of each request you make to DCS, including the [requestID and timestamp](https://dcs-pilot-docs.cloudapps.digital/sign-and-encrypt-a-DCS-payload/#build-a-plain-json-payload) of each request
```
PippaClarkGDS, line 16:
```suggestion
[Contact the DCS team](/support) to request access to production. You'll need to confirm your service:

+ is ready to use, for example, once you're connected to production someone could use it for the purpose set out in the use case you submitted as part of your DCS pilot application
+ has [successfully handled the 3 outcomes from the test requests](/send-a-test-request-to-the-DCS) you sent to the DCS integration environment (`valid: true`, `valid: false` and `error: true`)
+ has an auditing system set up to keep a record of each request you make to DCS, including the [requestID and timestamp](https://dcs-pilot-docs.cloudapps.digital/sign-and-encrypt-a-DCS-payload/#build-a-plain-json-payload) of each request
```

PippaClarkGDS (commented):
Hi Ben  - looks good, a few minor tweaks

PippaClarkGDS (approved):

PR #227, opened 2020-10-15
included a line about how scheduledOutages field will be empty for pilot participants

PippaClarkGDS:
## Why

Users getting confused when `scheduledOutages` field is empty

### DIFF
--- a/source/api-reference.html.md.erb
+++ b/source/api-reference.html.md.erb
@@ -43,7 +43,7 @@ If the DCS is available, the JSON response body will be similar to:
 }
 ```
 
-For the DCS pilot participants, the `scheduledOutages` field will be empty. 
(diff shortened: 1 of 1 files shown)


source/api-reference.html.md.erb
```diff
@@ -43,6 +43,8 @@ If the DCS is available, the JSON response body will be similar to:
 }
 ```
 
+For the DCS pilot participants, the `scheduledOutages` field will be empty.
```
chrisclayson, line 46: I wonder if this might lead to questions from the pilot participants about "why don't we get scheduled outage notifications" ... would it be better to say something like "this is legacy field that is now always empty, but provided for backward compatibility reasons for older clients"!??

PippaClarkGDS: cheers @chrisclayson  - that gives a bit more info. I've tweaked the order but kept the sentiment the same. Pushing now...

chrisclayson (approved):
👍

PR #235, opened 2020-11-16
moved detailed information about DCS responses into the API reference … and included a link to direct users there from the DCS responses page

PippaClarkGDS:
## Why

2 places in the docs explaining what a DCS response means is confusing for the user. 

## What

Moved the detail about the DCS responses to the API reference and included a link to direct users there from the more general overview in 'How the DCS works' 

## How to review

Sense check it makes sense for the detailed information about the DCS responses to be in the API reference


mrwilson (approved):

mrwilson:
Looks good to me @PippaClarkGDS

PR #240, opened 2020-12-02
Change homepage tenses from future to present

benjamin-mortimer:
The docs homepage references things that are currently happening as though they were going to happen in future. It also refers to "early 2020" as being in the future, but this date is now in the past.

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -2,7 +2,7 @@
 title: The Document Checking Service pilot
 weight: 1
 last_reviewed_on: 2020-12-02
-review_in: 3 months
+review_in: 6 months
 ---
 
 # The Document Checking Service pilot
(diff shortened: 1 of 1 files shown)


source/index.html.md.erb
```diff
@@ -1,15 +1,15 @@
 ---
 title: The Document Checking Service pilot
 weight: 1
-last_reviewed_on: 2020-10-19
+last_reviewed_on: 2020-12-02
 review_in: 3 months
```
mrwilson, line 5: Can we bump this to 6/9 months?

benjamin-mortimer: Done - I was thinking it would be good to comprehensively revisit this page in the near future but we should do that via our planning process rather than prompting the spaniel to remind us.

mrwilson (approved):

PR #259, opened 2021-03-01
BAU: Remove confusing link and references to JOSE

kerrr:
## Why

The document contains a link labeled "a JOSE object" that points to documentation about how to create a certificate thumbprint. This is confusing. We should change it.

## What

We do not have a document the simply explains what a JOSE object is, furthermore we rarely mention JOSE and can avoid doing so.  So we will remove all references to JOSE in our text. Technical uses in `Content-Type` and names of libraries remain, without comment.

## Technical writer support

Does this all make sense?

### DIFF
--- a/source/access-the-data-in-a-DCS-response.html.md.erb
+++ b/source/access-the-data-in-a-DCS-response.html.md.erb
@@ -10,7 +10,8 @@ review_in: 6 months
 When you make a successful request to the Document Checking Service (DCS), the DCS will respond with:
 
 * a `200` HTTP response code
-* a [JOSE object](/set-up-your-JOSE-certificate-thumbprints/#set-up-your-jose-certificate-thumbprints) containing the data you requested
+* a secured JSON object with data you requested
+
 
 Follow these instructions to access the data in the DCS response. You’ll need to:
 
--- a/source/sign-and-encrypt-a-DCS-payload.html.md.erb
+++ b/source/sign-and-encrypt-a-DCS-payload.html.md.erb
@@ -71,7 +71,7 @@ For the DCS, you need to set the following headers:
 1. `x5t#S256` contains the SHA256 thumbprint of your certificate.
 1. `alg` must be set to one of: `RS256`, `RS384`, `RS512`, `PS256`, `PS384`, or `PS512`.
 
-You must set the `x5t#S256` header. To set it, you’ll need to have [generated a thumbprint for your signing certificate](/set-up-your-JOSE-certificate-thumbprints).
+You must set the `x5t#S256` header. To set it, you’ll need to [generate a thumbprint](/generate-certificate-thumbprints) for your signing certificate.
 
 Depending on the library you use, you might have to manually construct the JWS header. Your header should look like this:
 
@@ -101,7 +101,7 @@ You must set the following headers:
 * `typ` must be set to `JWE`
 * `x5t#S256` contains the SHA256 thumbprint of the DCS encryption certificate
 
-You must set the `x5t#S256` header. To set it, you’ll need to have [generated a thumbprint for the DCS encryption certificate](/set-up-your-JOSE-certificate-thumbprints).
+You must set the `x5t#S256` header. To set it, you’ll need to [generate a thumbprint](/generate-certificate-thumbprints) for the DCS encryption certificate.
 
 Your header should look like this:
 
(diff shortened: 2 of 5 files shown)


source/access-the-data-in-a-DCS-response.html.md.erb
```diff
@@ -10,7 +10,7 @@ review_in: 6 months
 When you make a successful request to the Document Checking Service (DCS), the DCS will respond with:
 
 * a `200` HTTP response code
-* a [JOSE object](/Set-up-your-JOSE-certificate-thumbprints/#set-up-your-jose-certificate-thumbprints) containing the data you requested
+* a JSON object with data you requested, that has been secured using the [Sign and encrypt a DCS payload](/sign-and-encrypt-a-DCS-payload/#sign-and-encrypt-a-dcs-payload) process
```
benjamin-mortimer, line 13: I think it might be best just to remove the link.

I agree the current link is confusing.

However, the rationale for adding a link here appears to be to explain what a JOSE object is. Any content we link to that doesn't do that will be confusing, and we don't have any content that does it.

kerrr: Searching the docs for `JOSE` I see that it's only used:
 * when detailing the `Content-Type` request header, a technical detail
 * in the title of `Set up your JOSE certificate thumbprints`, instead of saying `JWE and JWS`
 * links referring to that doc
 * in names of libraries we point people to
 * in the `Sign and encrypt a DCS payload` document, in the `Using encryption algorithms` section when saying JWE would be better.

Maybe we should have an explainer on JOSE / JWT / JWE / JWS or avoid the term completely?

PippaClarkGDS: As a user, this might be an example of too deep for a level of technical understanding. E.g. when I get a new phone, I don't need it explaining how the processor works, I just need the setup info so I can do the thing. Agree that the link could likely confuse the user. 

Re the JOSE term used more widely in the docs, it's worth a conversation about whether it adds anything to the user (e.g. do they *need* to know about JOSE or can they just get on and do the thing without having to know)? We want to minimise cognitive load for them wherever possible.

kerrr: Yeah, I'm thinking that we should probably avoid using JOSE at all and just use "JWS", "JWE" and "JWS and JWE" where applicable. Possible with one mention and very short description when pointing to libraries, so that they know that JOSE is an encompassing term for al the related JWS, JWE, JWT technologies, so that's what the libs are branded with.

kerrr:
How about this update?

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -71,7 +71,7 @@ For the DCS, you need to set the following headers:
 1. `x5t#S256` contains the SHA256 thumbprint of your certificate.
 1. `alg` must be set to one of: `RS256`, `RS384`, `RS512`, `PS256`, `PS384`, or `PS512`.
 
-You must set the `x5t#S256` header. To set it, you’ll need to have [generated a thumbprint for your signing certificate](/set-up-your-JOSE-certificate-thumbprints).
+You must set the `x5t#S256` header. To set it, you’ll need to have [generated a thumbprint for your signing certificate](/set-up-your-JWS-and-JWE-certificate-thumbprints).
```
PippaClarkGDS, line 74:
```suggestion
You must set the `x5t#S256` header. To set it, you’ll need to have [generated JWS and JWE thumbprints for your signing certificate](/set-up-your-JWS-and-JWE-certificate-thumbprints).
```

source/sign-and-encrypt-a-DCS-payload.html.md.erb
```diff
@@ -101,7 +101,7 @@ You must set the following headers:
 * `typ` must be set to `JWE`
 * `x5t#S256` contains the SHA256 thumbprint of the DCS encryption certificate
 
-You must set the `x5t#S256` header. To set it, you’ll need to have [generated a thumbprint for the DCS encryption certificate](/set-up-your-JOSE-certificate-thumbprints).
+You must set the `x5t#S256` header. To set it, you’ll need to have [generated a thumbprint for the DCS encryption certificate](/set-up-your-JWS-and-JWE-certificate-thumbprints).
```
PippaClarkGDS, line 104:
```suggestion
You must set the `x5t#S256` header. To set it, you’ll need to have [generated JWS and JWE thumbprints for the DCS encryption certificate](/set-up-your-JWS-and-JWE-certificate-thumbprints).
```

PippaClarkGDS (approved):
Looks good - small tweak to make the link description more clear about thumbprints but otherwise all fine

source/sign-and-encrypt-a-DCS-payload.html.md.erb
kerrr: We're in a section that is talking about JWS headers, so the need the thumbprint of the signing certificate. I don't think it makes sense to say "... you’ll need to have generated JWS and JWE thumbprints for your signing certificate".

source/sign-and-encrypt-a-DCS-payload.html.md.erb
kerrr: Similar to above

kerrr:
As discussed with Pippa, made the certificate thumbprint doc a bit more generic.

PippaClarkGDS (approved):
Looks good Kerr 👍 !

